PR #29, opened 2020-11-11
Added script to install kustomize from source

murphybytes:
# Description

The kustomize script to install binaries was broken which in turn broke our Makefile.  The PR adds a script to install kustomize from source if kustomize is not already installed.  It has a number of advantages over the old way of doing things.

1. Tagged source history is almost always immutable, the likelihood that a breaking change will be introduced from the kustomize project is low. 
1. This script will not silently overwrite a user's existing kustomize install like the old install would. 
1. The criticism that this approach will increase build times is unwarranted.  While it is true that the install will increase build time, it's the increase would be no more than importing an additional package into the project which we do all the time without a second thought.  On a development machine, it's a one time 4 second hit. 

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,8 +53,7 @@ install-kubebuilder:
 
 .PHONY: install-kustomize
 install-kustomize:
-	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
-	mv kustomize ${KUSTOMIZE_INSTALL_DIR}/
+	@script/install-kustomize.sh
 
 # Uninstall CRDs from a cluster
 .PHONY: uninstall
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -53,8 +53,7 @@ install-kubebuilder:
 
 .PHONY: install-kustomize
 install-kustomize:
-	curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash -s 3.8.6
-	mv kustomize ${KUSTOMIZE_INSTALL_DIR}/
+	@script/install-kustomize.sh
```
aleksej-paschenko, line 56: if it does not take the destination directory, will [this](https://github.com/shipa-corp/ketch/blob/main/.github/workflows/release.yaml#L35) work?

murphybytes: Nope, good catch.  I'll fix it.

vivek:
This issue with kustomize is no longer present, ok to close this PR @murphybytes?

vivek:
Closing this as kustomize issue is not more present, we can revisit if it becomes issue.

PR #32, opened 2020-11-17
Ketch CI integration script

itsksaurabh:
# Description

### Usage
image

### Success Scenario 
image

## Type of change

- [x] Chore (documentation addition or typo, file relocation)

## Testing

## Documentation

## Final Checklist:


ci/deploy.sh
```diff
@@ -0,0 +1,96 @@
...
+
+# set defaults if not set by user
+if [ -z "$KETCH_TAG" ]; then 
+    KETCH_TAG=$(curl -s https://api.github.com/repos/shipa-corp/ketch/releases/latest | grep -Eo '"tag_name":.*[^\\]",' | head -n 1 | sed 's/[," ]//g' | cut -d ':' -f 2)
+fi
+
+# Install ketch binary at /usr/local/bin default location
+ curl -s https://raw.githubusercontent.com/shipa-corp/ketch/main/install.sh | TAG="${KETCH_TAG}" bash
+
+# Install Ketch controller
```
vivek, line 79: Should this script install ketch cluster on target k8s cluster? If at all it does, it should do if its either not installed or its a failed deployment.

ci/deploy.sh
```diff
@@ -0,0 +1,96 @@
...
+kubectl apply -f https://github.com/shipa-corp/ketch/releases/download/"${KETCH_TAG}"/ketch-controller.yaml
+
+if [ $RESOURCE_CREATION ] ; then
```
vivek: How am I supposed to use this condition, it's always true? you mean `RESOURCE_CREATION=false ./deploy ...` to skip pool/app creation? If so then the check should be:
```
if [[ $RESOURCE_CREATION == "true" ]] ; then
```

and RESOURCE_CREATION should be declared as:
```
: "${RESOURCE_CREATION:=true}"
```

ci/deploy.sh
```diff
@@ -0,0 +1,96 @@
...
+    if [ -z "$POOL" ]; then usage "Pool name required"; fi;
+    if [ -z "$INGRESS_ENDPOINT" ]; then usage "Ingress endpoint required"; fi;
+
+    # Add a pool with ingress Traefik (default), replace ingress endpoint address by your ingress IP address
+    echo "creating pool for deployment ..."
+    ketch pool add "${POOL}"  --ingress-service-endpoint "${INGRESS_ENDPOINT}" --ingress-type "${INGRESS_TYPE}"
+
+    # Create app
+    echo "creating app for deployment ..."
+    ketch app create "${APP_NAME}" --pool "${POOL}"  --env "${APP_ENV}" --registry-secret "${REG_SECRET}"
```
vivek, line 93: Also check for APP_NAME and print usage message, just like POOL and INGRESS_ENDPOINT?

ci/deploy.sh
```diff
@@ -0,0 +1,96 @@
...
+fi
+
+# Deploy app
+echo "deploying app ..."
+ketch app deploy "${APP_NAME}" -i "${DOCKER_REGISTRY}" --ketch-yaml "${KETCH_YAML}" --procfile "${PROCFILE}"
```
vivek: Add check for presence of DOCKER_REGISTRY?

vivek (changes requested):
@itsksaurabh Looks pretty good. Left some comments.

vivek:
About the name of script, should we use generic `deploy.sh` or specific to ci, `ketch-ci.sh`. Any other suggestions, @aleksej-paschenko @murphybytes @itsksaurabh?

ci/deploy.sh
```diff
@@ -0,0 +1,96 @@
...
+if [ -z "$DOCKER_IMAGE" ]; then usage "Image for the app is required"; fi;
+
+# set defaults if not set by user
+if [ -z "$KETCH_TAG" ]; then 
+    KETCH_TAG=$(curl -s https://api.github.com/repos/shipa-corp/ketch/releases/latest | grep -Eo '"tag_name":.*[^\\]",' | head -n 1 | sed 's/[," ]//g' | cut -d ':' -f 2)
+fi
+
+# Install ketch binary at /usr/local/bin default location
+ curl -s https://raw.githubusercontent.com/shipa-corp/ketch/main/install.sh | TAG="${KETCH_TAG}" bash
+
```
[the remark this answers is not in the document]
itsksaurabh: @vivek The ingress controller (Traefik or Istio), cluster issuer, and cert-manager can be pre-installed inside the cluster by the user. So, I skipped those installations as it can override old resources.

Please have a look at https://github.com/shipa-corp/ketch/pull/32/files#diff-794891fcb80950b0c81f330c416efdf5d35789fb56a91a18dcc25266d7cdccfdR5

ci/deploy.sh
```diff
@@ -0,0 +1,96 @@
...
+# Install Ketch controller
+kubectl apply -f https://github.com/shipa-corp/ketch/releases/download/"${KETCH_TAG}"/ketch-controller.yaml
```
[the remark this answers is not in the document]
itsksaurabh: I agree. ketch won't work without them.

ci/deploy.sh
itsksaurabh: We already do it [here](https://github.com/shipa-corp/ketch/pull/32/files#diff-d498a12aa774941c9553350f6bc684a7bb2a70d6b6a2c1fb0619dd90907e8d4dR68)

These checks ( POOL, INGRESS_ENDPOINT) are under the `if` block as they might not be required if --skip-resource-creation is not set.

ci/deploy.sh
itsksaurabh: We already do it [here](https://github.com/shipa-corp/ketch/pull/32/files#diff-d498a12aa774941c9553350f6bc684a7bb2a70d6b6a2c1fb0619dd90907e8d4dR69)

ci/deploy.sh
itsksaurabh: I agree. We need a check for it.

itsksaurabh:
> About the name of script, should we use generic `deploy.sh` or specific to ci, `ketch-ci.sh`. Any other suggestions, @aleksej-paschenko @murphybytes @itsksaurabh?

@vivek Yes. I was also thinking of it. Initially I named it `ketch-it.sh` hehe but changed afterward. `ketch-ci.sh` looks great to me. :))

ci/deploy.sh
vivek: ah, missed --skip-resource-creation flag. thanks.

ci/deploy.sh
```diff
@@ -0,0 +1,98 @@
...
+    ketch pool add "${POOL}"  --ingress-service-endpoint "${INGRESS_ENDPOINT}" --ingress-type "${INGRESS_TYPE}"
+
+    # Create app
+    echo "creating app for deployment ..."
+    ketch app create "${APP_NAME}" --pool "${POOL}"  --env "${APP_ENV}" --registry-secret "${REG_SECRET}"
+fi
+
+# Deploy app
+echo "deploying app ..."
+ketch app deploy "${APP_NAME}" -i "${DOCKER_IMAGE}" --ketch-yaml "${KETCH_YAML}" --procfile "${PROCFILE}"
```
vivek, line 98: It doesn't work for me: I tried:
`./deploy.sh -i docker.io/shipasoftware/bulletinboard:1.0  -a bb -o pool1 --endpoint 104.196.226.101`

It tries to run with ketch-yaml and procfile. These are optional. Maybe we want to build command string and include params if present? Something like:

```
echo "deploying app ..."
CMD_STR="ketch app deploy ${APP_NAME} -i ${DOCKER_IMAGE}"

if [ "${KETCH_YAML}" ]; then
  CMD_STR="${CMD_STR} --ketch-yaml ${KETCH_YAML}"
fi

if [ "${PROCFILE}" ]; then
  CMD_STR="${CMD_STR} --procfile ${PROCFILE}"
fi

echo "CMD: ${CMD_STR}"
eval "${CMD_STR}"
```

vivek (changes requested):

ci/deploy.sh
vivek: @itsksaurabh I think we should do `s/DOCKER_REGISTRY/DOCKER_IMAGE`?

ci/deploy.sh
itsksaurabh: It worked for me with the latest changes. But I completely agree with you that those commands should be built dynamically. There is no reason to keep all the flags if not required. Thanks!

ci/deploy.sh
itsksaurabh: @vivek, I liked this idea but I would like to do this validation inside Ketch.

 The reason for not including here is that different registries can have different checks for image:tag combination already exists. 

even for Dockerhub, we have to deal with two versions of the Docker Registry. For example:

for Docker registry V1,
```
function docker_tag_exists() {
    curl --silent -f -lSL https://index.docker.io/v1/repositories/$1/tags/$2 > /dev/null
}

if docker_tag_exists library/nginx 1.7.5; then
    echo exist
else 
    echo not exists
fi
```

In case of usage Docker Registry v2 :
```
# set username and password
UNAME="user"
UPASS="password"

function docker_tag_exists() {
    TOKEN=$(curl -s -H "Content-Type: application/json" -X POST -d '{"username": "'${UNAME}'", "password": "'${UPASS}'"}' https://hub.docker.com/v2/users/login/ | jq -r .token)
    curl --silent -f --head -lL https://hub.docker.com/v2/repositories/$1/tags/$2/ > /dev/null
}

if docker_tag_exists library/nginx 1.7.5; then
    echo exist
else 
    echo not exists
fi
```
non-docker.io registries will have different ways to do it. 
So, I think we should do it with ketch otherwise this script will have tons of unused commands. Thanks!

murphybytes (approved):

PR #43, opened 2020-12-17
remove namespace with pool remove command

parkerduckworth:
# Description

Kubebuilder does not remove namespaces with the deletion of Kubernetes objects, even if the namespace is empty.  This PR introduces three additional client calls to the `poolRemove` function:

1. Get the pool-to-be-removed so we have access to its namespace (only have pool name with cli args)
2. Get the namespace associated with the pool-to-be-removed
3. Remove the associated namespace

## Type of change

## Testing

While I did write a test, I did not include it here.  The [fake client](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client/fake#ClientBuilder) included with Kubebuilder does not alter cluster namespace when creating a pool, so `poolRemove` fails with `failed to get namespace`.  I had a test that ignores this error, but it defeats the purpose of testing IMO so I left it out.

I did come across [envtest](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest#Environment.Start) which may enable us to test for namespace removal.  How do we feel about bringing this into our test suite? It is quite a bit more involved, setting up a full test environment + API Server.

With that said, I did test these changes locally and getting the namespace name from the pool lookup to remove the namespace works as expected, terminating the namespace associated with the pool-to-be-removed.

## Documentation

## Final Checklist:

### DIFF
--- a/cmd/ketch/pool_remove.go
+++ b/cmd/ketch/pool_remove.go
@@ -52,12 +52,37 @@ func poolRemove(ctx context.Context, cfg config, options poolRemoveOptions, out 
 		return fmt.Errorf("failed to remove the pool: %w", err)
 	}
 
-	// Remove namespace after pool is deleted
-	if err := cfg.Client().Delete(ctx, &ns); err != nil {
-		return fmt.Errorf("failed to remove the namespace: %w", err)
+	if userWantsToRemoveNamespace(ns.Name, out) {
+		if err := cfg.Client().Delete(ctx, &ns); err != nil {
+			return fmt.Errorf("failed to remove the namespace: %w", err)
+		}
+		fmt.Fprintln(out, "Namespace successfully removed!")
 	}
 
-	fmt.Fprintln(out, "Successfully removed!")
+	fmt.Fprintln(out, "Pool successfully removed!")
 
 	return nil
+}
+
+func userWantsToRemoveNamespace(ns string, out io.Writer) bool {
+	response := promptToRemoveNamespace(ns, out)
+	return handleNamespaceRemovalResponse(response, ns, out)
+}
+
+func promptToRemoveNamespace(ns string, out io.Writer) string {
+	fmt.Fprintf(out, "Do you want to remove the namespace along with the pool? Please enter namespace to confirm (%s): ", ns)
+
+	var response string
+	fmt.Scanln(&response)
+
+	return response
+}
+
+func handleNamespaceRemovalResponse(response, ns string, out io.Writer) bool {
+	if response != ns {
+		fmt.Fprintln(out, "Skipping namespace removal...")
+		return false
+	}
+
+	return true
 }
--- a/cmd/ketch/pool_remove.go
+++ b/cmd/ketch/pool_remove.go
@@ -115,7 +115,7 @@ func removeNamespace(ctx context.Context, cfg config, pool *ketchv1.Pool) error 
 	var ns corev1.Namespace
 
 	if err := cfg.Client().Get(ctx, types.NamespacedName{Name: pool.Spec.NamespaceName}, &ns); err != nil {
-		return fmt.Errorf("failed to get namespace: %s", err)
+		return fmt.Errorf("failed to get namespace: %w", err)
 	}
 
 	if err := cfg.Client().Delete(ctx, &ns); err != nil {
--- a/cmd/ketch/pool_remove.go
+++ b/cmd/ketch/pool_remove.go
@@ -6,7 +6,8 @@ import (
 	"io"
 
 	"github.com/spf13/cobra"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 
 	ketchv1 "github.com/shipa-corp/ketch/internal/api/v1beta1"
 )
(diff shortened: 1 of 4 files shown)


cmd/ketch/pool_remove.go
```diff
@@ -35,14 +36,28 @@ type poolRemoveOptions struct {
 }
 
 func poolRemove(ctx context.Context, cfg config, options poolRemoveOptions, out io.Writer) error {
```
itsksaurabh, line 38: can we have a test for it?

parkerduckworth: While I did write a test, I did not include it here.  The [fake client](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client/fake#ClientBuilder) included with Kubebuilder does not alter cluster namespace when creating a pool, so `poolRemove` fails with `failed to get namespace`.  I had a test that ignores this error, but it defeats the purpose of testing IMO so I left it out.

I did come across [envtest](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/envtest#Environment.Start) which may enable us to test for namespace removal.  How do we feel about bringing this into our test suite? It is quite a bit more involved, setting up a full test environment + API Server.

With that said, I did test these changes locally and getting the namespace name from the pool lookup to remove the namespace works as expected, terminating the namespace associated with the pool-to-be-removed.

cmd/ketch/pool_remove.go
```diff
@@ -35,14 +36,28 @@ type poolRemoveOptions struct {
...
+	if err := cfg.Client().Get(ctx, types.NamespacedName{Name: options.Name}, &pool); err != nil {
+		return fmt.Errorf("failed to get pool: %w", err)
 	}
+
+	// Get namespace that was created with the pool so it can be removed
+	if err := cfg.Client().Get(ctx, types.NamespacedName{Name: pool.Spec.NamespaceName}, &ns); err != nil {
+		return fmt.Errorf("failed to get namespace: %w", err)
+	}
+
 	if err := cfg.Client().Delete(ctx, &pool); err != nil {
```
itsksaurabh, line 51: We already pass pool in `cfg.Client().Delete` func. Is it possible to put namespace removal logic inside it instead of here?

cmd/ketch/pool_remove.go
itsksaurabh: Could you please try k8s.io/client-go/kubernetes/fake ?

cmd/ketch/pool_remove.go
parkerduckworth: `cfg.Client().Delete` is a function from a dependency, the Kubebuilder controller-runtime library 🙂 

https://github.com/kubernetes-sigs/controller-runtime/blob/cf7ac887d082b481ec1389ddcd576f4140b15595/pkg/client/client.go#L167

Unless I misunderstood?

cmd/ketch/pool_remove.go
parkerduckworth: I already wrote a test file using fake, and when it runs,  `poolRemove` fails with `failed to get namespace` because `fake` doesn't create a namespace when the test creates a pool.  I can push the file to this PR if you want to take a look.

cmd/ketch/pool_remove.go
```diff
@@ -35,14 +36,28 @@ type poolRemoveOptions struct {
...
 		return fmt.Errorf("failed to remove the pool: %w", err)
 	}
+
+	// Remove namespace after pool is deleted
+	if err := cfg.Client().Delete(ctx, &ns); err != nil {
```
murphybytes, line 56: IMO we should prompt the user for confirmation to delete the namespace associated with the pool.  They may be using the namespace for something else.  I'd ask the user to confirm removal of the namespace by typing out the name of the namespace if they actually want to delete it.

murphybytes (changes requested):

cmd/ketch/pool_remove.go
parkerduckworth: Good idea 👍

aleksej-paschenko: agree, that's a good idea
just understand that I usually use "default" namespace for ketch pool

cmd/ketch/pool_remove.go
parkerduckworth: @itsksaurabh test has been fixed with commit [47fe2f12](https://github.com/shipa-corp/ketch/pull/43/commits/47fe2f12b63853544a9288302ec1443b77f72de6)

murphybytes (commented):
What if other pools use this namespace?  I think we should prevent the user from removing in this case, but I'd be curious about what others think?

cmd/ketch/pool_remove.go
```diff
@@ -35,14 +43,84 @@ type poolRemoveOptions struct {
...
+	}
+
+	return nil
+}
+
+func removeNamespace(ctx context.Context, cfg config, pool *ketchv1.Pool) error {
+	var ns corev1.Namespace
+
+	if err := cfg.Client().Get(ctx, types.NamespacedName{Name: pool.Spec.NamespaceName}, &ns); err != nil {
+		return fmt.Errorf("failed to get namespace: %s", err)
```
murphybytes, line 118: You should wrap the returned error with the %w verb instead of %s

murphybytes (changes requested):
This looks great! I like how you make it really easy to understand what's going on with the code.  I caught one tiny mistake, fix that and I'll approve.

cmd/ketch/pool_remove.go
```diff
@@ -35,14 +43,84 @@ type poolRemoveOptions struct {
...
+func handleNamespaceRemovalResponse(response, ns string, out io.Writer) bool {
+	if response != ns {
+		fmt.Fprintln(out, skipNsRemovalMsg)
+		return false
+	}
+
+	return true
+}
+
+func namespaceHasAdditionalResources(ctx context.Context, cfg config, targetPool *ketchv1.Pool) error {
```
itsksaurabh, line 96: `namespaceHasAdditionalPools` ?

itsksaurabh: also, the function name sounds like it will return a bool. I would do something like `checkNamespaceAdditionalPools` or similar.

cmd/ketch/pool_remove.go
```diff
@@ -35,14 +43,84 @@ type poolRemoveOptions struct {
...
+	return nil
+}
+
+func userWantsToRemoveNamespace(ns string, out io.Writer) bool {
+	response := promptToRemoveNamespace(ns, out)
+	return handleNamespaceRemovalResponse(response, ns, out)
+}
+
+func promptToRemoveNamespace(ns string, out io.Writer) string {
+	fmt.Fprintf(out, "Do you want to remove the namespace along with the pool? Please enter namespace to confirm (%s): ", ns)
```
itsksaurabh, line 79: Do we check if the namespace provided exists?

cmd/ketch/pool_remove.go
```diff
@@ -6,14 +6,22 @@ import (
...
-const poolRemoveHelp = `
+const (
+	poolRemoveHelp = `
 Remove an existing pool.
 `
+	skipNsRemovalMsg = "Skipping namespace removal..."
+)
+
+var (
+	printNsRemovalErr = func(out io.Writer, err error) { fmt.Fprintf(out, "%s\n%s", err, skipNsRemovalMsg) }
```
itsksaurabh, line 23: Why do we need to store the anonymous func in a variable? Do we change need to override `printNsRemovalErr` somewhere? is it possible to make it a standard function ?

cmd/ketch/pool_remove.go
parkerduckworth: The namespace is extracted from the given pool name

cmd/ketch/pool_remove.go
parkerduckworth: Sure

cmd/ketch/pool_remove.go
parkerduckworth: 👍

murphybytes (approved):

PR #45, opened 2020-12-22
Add app log command to stream logs of an application

aleksej-paschenko:
This PR adds `ketch app log` command to stream logs of an application

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/cmd/ketch/app_log.go
+++ b/cmd/ketch/app_log.go
@@ -139,7 +139,7 @@ func watchLogs(cli kubernetes.Interface, options watchOptions, readLogs readLogs
 	if err != nil {
 		return err
 	}
-	// we are going to read logs from all running pods, just read without streaming
+	// we are going to read logs from all running pods, just read without streaming.
 	msgChs := make(map[types.UID]chan logMessage, len(pods.Items))
 	for _, pod := range pods.Items {
 		containerName, err := ketchContainerName(pod)
@@ -162,6 +162,7 @@ func watchLogs(cli kubernetes.Interface, options watchOptions, readLogs readLogs
 	}
 
 	// we need a timestamp of the last message of each pod, so later we will stream logs from this time.
+	// we avoid using sort.Sort because downloading all logs and keeping them in memory can be resource-consuming operation.
 	timeOfLastMessage := make(map[types.UID]time.Time)
 	for {
 		// on each iteration we are looking for a message with minimal time
--- a/cmd/ketch/app_log.go
+++ b/cmd/ketch/app_log.go
@@ -48,7 +48,7 @@ func newAppLogCmd(cfg config, out io.Writer, appLog appLogFn) *cobra.Command {
 	cmd.Flags().StringVarP(&options.processName, "process", "p", "", "Process name")
 	cmd.Flags().IntVarP(&options.deploymentVersion, "version", "v", 0, "Deployment version")
 	cmd.Flags().BoolVarP(&options.follow, "follow", "f", false, "Specify if the logs should be streamed")
-	cmd.Flags().BoolVar(&options.ignoreErrors, "ignore-errors", false, "if watching / following pod logs, allow for any errors that occur to be non-fatal")
+	cmd.Flags().BoolVar(&options.ignoreErrors, "ignore-errors", false, "If watching / following pod logs, allow for any errors that occur to be non-fatal")
 	cmd.Flags().BoolVar(&options.prefix, "prefix", false, "Prefix each log line with the log source (pod name and container name)")
 	cmd.Flags().BoolVar(&options.timestamps, "timestamps", false, "Include timestamps on each line in the log output")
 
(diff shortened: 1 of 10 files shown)


murphybytes (approved):
This looks great!  The only suggestions I had that I think with this type of long running operation its going to be a good idea to use a cancel context to pass through all the goroutines and calls to k8s client cleanly shut things down.  We'd cancel on SIGINT or SIGTERM. I think this approach would also help that cluster side resources are cleaned up faster.

murphybytes:
Oh, also needs tests.

cmd/ketch/app_log.go
```diff
@@ -0,0 +1,364 @@
...
+			if !validation.ValidateName(options.appName) {
+				return ErrInvalidAppName
+			}
+			return appLog(cmd.Context(), cfg, options, out, watchLogs)
+		},
+	}
+	cmd.Flags().StringVarP(&options.processName, "process", "p", "", "Process name")
+	cmd.Flags().IntVarP(&options.deploymentVersion, "version", "v", 0, "Deployment version")
+	cmd.Flags().BoolVarP(&options.follow, "follow", "f", false, "Specify if the logs should be streamed")
+	cmd.Flags().BoolVar(&options.ignoreErrors, "ignore-errors", false, "if watching / following pod logs, allow for any errors that occur to be non-fatal")
```
itsksaurabh, line 51: please start description with upper case as other flags

cmd/ketch/app_log.go
```diff
@@ -0,0 +1,364 @@
...
+		}
+		msgChs[pod.UID] = readLogs(cli.CoreV1().Pods(pod.Namespace).GetLogs, pod, *containerName, options.out)
+	}
+
+	// we want to show the logs sorted by timestamp.
+	// lets store one message per pod and then in a loop below we will select one message with minimal time on each iteration.
+	messages := make(map[types.UID]logMessage, len(msgChs))
+	for podUID, msg := range msgChs {
+		if m, ok := <-msg; ok {
+			messages[podUID] = m
```
itsksaurabh, line 157: not sure but, could we set it in the loop above?
https://github.com/shipa-corp/ketch/pull/45/files#diff-80131f1268c6db0851cfbef6ae1ef9c7d4b8f7e5ed98ffac81ffd8cf7c47707fR144

cmd/ketch/app_log.go
aleksej-paschenko: fixed, thanks!

cmd/ketch/app_log.go
aleksej-paschenko: yep, though it's possible I would keep it as is. The first loop runs several goroutines, each of them opens a connection to read logs. This gives a small performance boost)

cmd/ketch/testdata/app-log/1.output
```diff
@@ -0,0 +1,16 @@
+hello-web-1 0
```
itsksaurabh, line 1: The timestamp is an essential piece of information that should be included in every log entry. It's no use to know that something happened without knowing when it happened.
https://www.scalyr.com/blog/log-formatting-best-practices-readable/#:~:text=The%20timestamp%20is%20an%20essential,without%20knowing%20when%20it%20happened.

cmd/ketch/testdata/app-log/3.output
```diff
@@ -0,0 +1,16 @@
+[hello-web-1-random/hello-web-1] 2021-01-13 16:49:00 +0000 UTC hello-web-1 0
+[dashboard-worker-2-random/dashboard-worker-2] 2021-01-13 16:49:01 +0000 UTC dashboard-worker-2 0
```
itsksaurabh, line 2: Do you think it would be nice to start log with timestamp?
As a user it would be more convenient to me.

cmd/ketch/testdata/app-log/stream-1.output
```diff
@@ -0,0 +1,16 @@
+[hello-web-1-random/hello-web-1] 2021-01-13 16:49:00 +0000 UTC hello-web-1 0
+[hello-web-2-random/hello-web-2] 2021-01-13 16:49:01 +0000 UTC hello-web-2 0
+[hello-web-1-random/hello-web-1] 2021-01-13 16:49:05 +0000 UTC hello-web-1 1
+[hello-web-2-random/hello-web-2] 2021-01-13 16:49:06 +0000 UTC hello-web-2 1
```
itsksaurabh, line 4: same here

cmd/ketch/testdata/app-log/3.output
aleksej-paschenko: we are using the same format as `kubectl logs --prefix --timestamps`

cmd/ketch/testdata/app-log/1.output
aleksej-paschenko: it's optional, you can enable timestamps with `--timestamps`

itsksaurabh (approved):
LGTM

PR #47, opened 2020-12-25
[SHIPA-814] app status

c12o16h1:
# Description

Updates output of `app list` and `app info` commands, replacing column `UNITS` with `STATUS`.

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/cmd/ketch/app.go
+++ b/cmd/ketch/app.go
@@ -1,9 +1,14 @@
 package main
 
 import (
+	"fmt"
 	"io"
+	"strings"
 
 	"github.com/spf13/cobra"
+	apiv1 "k8s.io/api/core/v1"
+
+	ketchv1 "github.com/shipa-corp/ketch/internal/api/v1beta1"
 )
 
 func newAppCmd(cfg config, out io.Writer) *cobra.Command {
@@ -25,4 +30,81 @@ func newAppCmd(cfg config, out io.Writer) *cobra.Command {
 	cmd.AddCommand(newAppStopCmd(cfg, out, appStop))
 	cmd.AddCommand(newAppExportCmd(cfg, out, appExport))
 	return cmd
+}
+
+func appState(pods []apiv1.Pod) (state string) {
+	state = "unknown"
+	// If there's no pods for this app, then app in `created` state
+	if len(pods) == 0 {
+		state = strings.ToLower(string(ketchv1.AppCreated))
+		return state
+	}
+
+	// If app has pods, then we build app status based on pods statuses
+	var deploying, running, errored, succeeded int
+	for _, pod := range pods {
+		switch podState(pod) {
+		case ketchv1.PodRunning:
+			running++
+		case ketchv1.PodDeploying:
+			deploying++
+		case ketchv1.PodError:
+			errored++
+		case ketchv1.PodSucceeded:
+			succeeded++
+		}
+	}
+
+	var parts []string
+	if deploying > 0 {
+		parts = append(parts, fmt.Sprintf(`%d %s`, deploying, ketchv1.PodDeploying))
+	}
+	if running > 0 {
+		parts = append(parts, fmt.Sprintf(`%d %s`, running, ketchv1.PodRunning))
+	}
+	if errored > 0 {
+		parts = append(parts, fmt.Sprintf(`%d %s`, errored, ketchv1.PodError))
+	}
+	if succeeded > 0 {
+		parts = append(parts, fmt.Sprintf(`%d %s`, succeeded, ketchv1.PodSucceeded))
+	}
+
+	return strings.Join(parts, ", ")
+}
+
+var (
+	containerStateCrashLoopBackOff = "CrashLoopBackOff"
+	containerStateCompleted        = "Completed"
+)
+
+func podState(pod apiv1.Pod) ketchv1.PodState {
+	switch pod.Status.Phase {
+	case apiv1.PodPending:
+		return ketchv1.PodDeploying
+	case apiv1.PodRunning:
+		// In some cases, we have to override the status obtained by using `pod.Status.Phase` because it could be
+		// `running` state while the actual container in `CrashLoopBackOff` so, actual state of the pod is `error`,
+		// and so on.
+		for _, c := range pod.Status.ContainerStatuses {
+			// If any of the pod containers in the `CrashLoopBackOff` state we must return the `error` state
+			if c.State.Waiting != nil {
+				if c.State.Waiting.Reason == containerStateCrashLoopBackOff {
+					return ketchv1.PodError
+				}
+			}
+			// If any of the pod containers in the `Completed` state we must return the `succeeded` state
+			if c.State.Terminated != nil {
+				if c.State.Terminated.Reason == containerStateCompleted {
+					return ketchv1.PodSucceeded
+				}
+			}
+		}
+		return ketchv1.PodRunning
+	case apiv1.PodFailed, apiv1.PodUnknown:
+		return ketchv1.PodError
+	case apiv1.PodSucceeded:
+		return ketchv1.PodSucceeded
+	default:
+		return ketchv1.PodError
+	}
 }
--- a/cmd/ketch/app_info.go
+++ b/cmd/ketch/app_info.go
@@ -10,6 +10,7 @@ import (
 	"text/template"
 
 	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	ketchv1 "github.com/shipa-corp/ketch/internal/api/v1beta1"
@@ -85,24 +86,30 @@ func appInfo(ctx context.Context, cfg config, options appInfoOptions, out io.Wri
 	if err := cfg.Client().Get(ctx, types.NamespacedName{Name: app.Spec.Pool}, pool); err != nil {
 		return fmt.Errorf("failed to get pool: %w", err)
 	}
+
 	buf := bytes.Buffer{}
 	t := template.Must(template.New("app-info").Parse(appInfoTemplate))
 	table := &bytes.Buffer{}
 	w := tabwriter.NewWriter(table, 0, 4, 4, ' ', 0)
-	fmt.Fprintln(w, "DEPLOYMENT VERSION\tIMAGE\tPROCESS NAME\t#UNITS\tCMD")
+	fmt.Fprintln(w, "DEPLOYMENT VERSION\tIMAGE\tPROCESS NAME\tSTATE\tCMD")
 	noProcesses := true
 	for _, deployment := range app.Spec.Deployments {
 		for _, process := range deployment.Processes {
 			noProcesses = false
-			units := fmt.Sprintf("%d", ketchv1.DefaultNumberOfUnits)
-			if process.Units != nil {
-				units = fmt.Sprintf("%d", *process.Units)
+
+			pods, err := cfg.KubernetesClient().CoreV1().Pods(app.Namespace).List(ctx, metav1.ListOptions{
+				LabelSelector: fmt.Sprintf(`theketch.io/app-name=%s,theketch.io/app-deployment-version=%s,theketch.io/app-process=%s`, app.Name, deployment.Version, process.Name),
+			})
+			if err != nil {
+				return err
 			}
+			state := appState(pods.Items)
+
 			line := []string{
 				deployment.Version.String(),
 				deployment.Image,
 				process.Name,
-				units,
+				state,
 				strings.Join(process.Cmd, " "),
 			}
 			fmt.Fprintln(w, strings.Join(line, "\t"))
--- a/cmd/ketch/app_info.go
+++ b/cmd/ketch/app_info.go
@@ -129,30 +129,13 @@ func appInfo(ctx context.Context, cfg config, options appInfoOptions, out io.Wri
 	return nil
 }
 
-func filterProcessDeploymentPods(appPods []corev1.Pod, deployment, process string) []corev1.Pod {
+func filterProcessDeploymentPods(appPods []corev1.Pod, version, process string) []corev1.Pod {
 	var pods []corev1.Pod
-Loop:
 	for _, pod := range appPods {
-		match := 0
-		desired := 2
-		for label, value := range pod.GetLabels() {
-			if label == ketchDeploymentVersionLabel {
-				if value != deployment {
-					continue Loop
-				}
-				match++
-			}
-			if label == ketchProcessNameLabel {
-				if value != process {
-					continue Loop
-				}
-				match++
-			}
-
-			if match == desired {
-				pods = append(pods, pod)
-				continue Loop
-			}
+		deploymentVersion := pod.Labels[ketchDeploymentVersionLabel]
+		processName := pod.Labels[ketchProcessNameLabel]
+		if deploymentVersion == version && processName == process {
+			pods = append(pods, pod)
 		}
 	}
 	return pods
--- a/cmd/ketch/app_info.go
+++ b/cmd/ketch/app_info.go
@@ -10,6 +10,7 @@ import (
 	"text/template"
 
 	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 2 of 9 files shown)


cmd/ketch/app.go
```diff
@@ -26,3 +31,80 @@ func newAppCmd(cfg config, out io.Writer) *cobra.Command {
 	cmd.AddCommand(newAppExportCmd(cfg, out, appExport))
 	return cmd
 }
+
+func appState(pods []apiv1.Pod) (state string) {
```
aleksej-paschenko, line 35: Let's use a more structured approach, 

```golang
type AppState struct {
    RunningPods int
    ErrorrPods int
    DeployingPods int
}
func appState(pods []apiv1.Pod) AppState {

}

func (s AppState) String() string {
   
}
```

cmd/ketch/app_info.go
```diff
@@ -85,24 +86,30 @@ func appInfo(ctx context.Context, cfg config, options appInfoOptions, out io.Wri
...
 	noProcesses := true
 	for _, deployment := range app.Spec.Deployments {
 		for _, process := range deployment.Processes {
 			noProcesses = false
-			units := fmt.Sprintf("%d", ketchv1.DefaultNumberOfUnits)
-			if process.Units != nil {
-				units = fmt.Sprintf("%d", *process.Units)
+
+			pods, err := cfg.KubernetesClient().CoreV1().Pods(app.Namespace).List(ctx, metav1.ListOptions{
+				LabelSelector: fmt.Sprintf(`theketch.io/app-name=%s,theketch.io/app-deployment-version=%s,theketch.io/app-process=%s`, app.Name, deployment.Version, process.Name),
```
aleksej-paschenko, line 101: it would be helpful to put labels to "labels.go", 
could you take it from 
https://github.com/shipa-corp/ketch/pull/45/files#diff-59fb47386f112a9a44e9e7fcf9594bf3346a4afa9ae47a170a9c20d1e3752bc6R4?

also, this pattern of querying kubernetes doesn't scale well, 
can we get all pods of all ketch applications before these loops and group them by an app, process, and deployment?

cmd/ketch/app_list.go
```diff
@@ -43,11 +44,19 @@ func appList(ctx context.Context, cfg config, out io.Writer) error {
 		poolsByName[pool.Name] = pool
 	}
 	w := tabwriter.NewWriter(out, 0, 4, 4, ' ', 0)
-	fmt.Fprintln(w, "NAME\tPOOL\tUNITS\tADDRESSES\tDESCRIPTION")
+	fmt.Fprintln(w, "NAME\tPOOL\tSTATE\tADDRESSES\tDESCRIPTION")
 	for _, item := range apps.Items {
+		pods, err := cfg.KubernetesClient().CoreV1().Pods(item.Namespace).List(ctx, metav1.ListOptions{
```
aleksej-paschenko, line 49: let's get pods of all ketch apps in one call and group them by an app.

cmd/ketch/app.go
```diff
@@ -26,3 +31,80 @@ func newAppCmd(cfg config, out io.Writer) *cobra.Command {
...
+		for _, c := range pod.Status.ContainerStatuses {
+			// If any of the pod containers in the `CrashLoopBackOff` state we must return the `error` state
+			if c.State.Waiting != nil {
+				if c.State.Waiting.Reason == containerStateCrashLoopBackOff {
+					return ketchv1.PodError
+				}
+			}
+			// If any of the pod containers in the `Completed` state we must return the `succeeded` state
+			if c.State.Terminated != nil {
+				if c.State.Terminated.Reason == containerStateCompleted {
```
aleksej-paschenko, line 97: what if there are two containers: the first one has `completed` state and the second has `containerStateCrashLoopBackOff` state?

cmd/ketch/app.go
```diff
@@ -26,3 +31,80 @@ func newAppCmd(cfg config, out io.Writer) *cobra.Command {
...
+
+	return strings.Join(parts, ", ")
+}
+
+var (
+	containerStateCrashLoopBackOff = "CrashLoopBackOff"
+	containerStateCompleted        = "Completed"
+)
+
+func podState(pod apiv1.Pod) ketchv1.PodState {
```
aleksej-paschenko, line 80: I think we should discuss this idea of having `ketchv1.PodState`, it is a refined version of `apiv1.PodPhase`. 
A summary of the refinements is:
1) rename PodPending to PodDeploying
2) rename PodFailed to PodError 
3) override a pod's PodRunning if something has happened with one of the pod's container. 

Do we need 1 and 2? 
People got used to kubernetes Pod State.

internal/api/v1beta1/app_types.go
```diff
@@ -497,3 +497,18 @@ func (s AppStatus) Condition(t AppConditionType) *AppCondition {
 	}
 	return nil
 }
+
+// PodState describes the simplified state of a pod in the cluster
+type PodState string
```
aleksej-paschenko, line 502: I would not introduce another pod state type, even internally. It would be confusing.

aleksej-paschenko (changes requested):

cmd/ketch/app_info.go
c12o16h1: Well, had same idea initially, but we can't get all pods for all apps, because apps could be in different namespaces. So, we need to do it at least for each app. Then, if we already have to do so per each app - there's not much improvement with getting all pods per app, because, usually - there's just only one deployment per app, and one process. Sure, there could be more in some rare cases, but overall, code readability will be worse with this patch. It not worth it.

cmd/ketch/app.go
c12o16h1: Didn't get what is the profit? More code, more complexity in code, same result. Why we need to complicate that?

cmd/ketch/app_list.go
c12o16h1: namespaces....

cmd/ketch/app_info.go
aleksej-paschenko: https://github.com/datawire/ambassador/blob/master/pkg/k8s/client.go#L36

cmd/ketch/app_list.go
aleksej-paschenko: https://github.com/datawire/ambassador/blob/master/pkg/k8s/client.go#L36

cmd/ketch/app.go
c12o16h1: then first happened error will take precedence. If we will want to improve that logic - let's do that then.

cmd/ketch/app.go
c12o16h1: Well, had the same question in my mind, this is the product requirements, not mine

internal/api/v1beta1/app_types.go
c12o16h1: Well, yes, but as this is not a direct translation of apiv1.PodPhase, and we have this overridden in some cases - seems that we don't have another choice than having separate states.

cmd/ketch/app.go
aleksej-paschenko: at least we can implement something like

```golang

func refinePodState(pod apiv1.Pod) apiv1.PodPhase {
	switch pod.Status.Phase {
	case apiv1.PodPending:
		return apiv1.PodPending
	case apiv1.PodRunning:
		// In some cases, we have to override the status obtained by using `pod.Status.Phase` because it could be
		// `running` state while the actual container in `CrashLoopBackOff` so, actual state of the pod is `error`,
		// and so on.
		for _, c := range pod.Status.ContainerStatuses {
			// If any of the pod containers in the `CrashLoopBackOff` state we must return the `error` state
			if c.State.Waiting != nil {
				if c.State.Waiting.Reason == containerStateCrashLoopBackOff {
					return apiv1.PodFailed
				}
			}
			// If any of the pod containers in the `Completed` state we must return the `succeeded` state
			if c.State.Terminated != nil {
				if c.State.Terminated.Reason == containerStateCompleted {
					return apiv1.PodSucceeded
				}
			}
		}
		return apiv1.PodRunning
	case apiv1.PodFailed, apiv1.PodUnknown:
		return apiv1.PodFailed
	case apiv1.PodSucceeded:
		return apiv1.PodSucceeded
	default:
		return apiv1.PodFailed
	}
}

```
but I suggest discussing the task's requirements

cmd/ketch/app_info.go
```diff
@@ -121,3 +128,32 @@ func appInfo(ctx context.Context, cfg config, options appInfoOptions, out io.Wri
 	fmt.Fprintf(out, "%v", buf.String())
 	return nil
 }
+
+func filterProcessDeploymentPods(appPods []corev1.Pod, deployment, process string) []corev1.Pod {
+	var pods []corev1.Pod
+Loop:
```
aleksej-paschenko, line 134: `Loop:` is a kind of BFG in golang, If you know what I mean 

how about a more straightforward approach: 
```golang
func filterProcessDeploymentPods(appPods []corev1.Pod, deployment, process string) []corev1.Pod {
	var pods []corev1.Pod
	for _, pod := range appPods {
		version := pod.Labels[ketchDeploymentVersionLabel]
		processName := pod.Labels[ketchProcessNameLabel]
		if version == deployment && processName == process {
			pods = append(pods, pod)
		}
	}
	return pods
}

```
or we can group all pods by deployment and process before using them
```golang

// Pods of the application grouped by a deployment version and then by a process name
pods := make(map[string]map[string][]corev1.Pod)

for _, pod := range appPods.Items {
	version := pod.Labels[ketchDeploymentVersionLabel]
	byVersion, ok := pods[version]
	if !ok {
		byVersion = map[string][]corev1.Pod{}
		pods[version] = byVersion
	}
	processName := pod.Labels[ketchProcessNameLabel]
	byVersion[processName] = append(byVersion[processName], pod)
}


```

cmd/ketch/app_list.go
```diff
@@ -42,14 +44,61 @@ func appList(ctx context.Context, cfg config, out io.Writer) error {
...
+	}
+
+	return cfg.KubernetesClient().CoreV1().Pods(metav1.NamespaceAll).List(ctx, metav1.ListOptions{
+		LabelSelector: s.String(),
+	})
+}
+
+func filterAppPods(appName string, pods []corev1.Pod) []corev1.Pod {
+	var appPods []corev1.Pod
+Loop:
```
aleksej-paschenko, line 94: again, no need to iterate over the label list and no need to use `Loop:`

cmd/ketch/app_update.go
```diff
@@ -45,6 +52,16 @@ type appUpdateOptions struct {
...
 	dockerRegistrySecretSet bool
 	dockerRegistrySecret    string
+	platformRemove          bool
+	platformSet             bool
+	platform                string
+}
+
+func (au appUpdateOptions) validate() error {
+	if au.platformRemove && au.platformSet {
+		return errors.New("Ambiguous platform commands.  You can't remove a platform and change a platform at the same time")
```
aleksej-paschenko, line 62: it looks like lines from another commit

cmd/ketch/app_update.go
```diff
@@ -56,6 +73,15 @@ func appUpdate(ctx context.Context, cfg config, options appUpdateOptions, out io
 	if err = cfg.Client().Get(ctx, types.NamespacedName{Name: options.name}, &app); err != nil {
 		return fmt.Errorf("failed to get app: %w", err)
 	}
+	if options.platformRemove {
+		app.Spec.Platform = ""
+	}
+	if options.platformSet {
```
aleksej-paschenko, line 79: same here

cmd/ketch/app_list.go
```diff
@@ -42,14 +44,61 @@ func appList(ctx context.Context, cfg config, out io.Writer) error {
...
 	w.Flush()
 	return nil
 }
+
+func allAppsPods(ctx context.Context, cfg config, apps []ketchv1.App) (*corev1.PodList, error) {
+	var appNames []string
+	for _, a := range apps {
+		appNames = append(appNames, a.Name)
+	}
+	if len(appNames) == 0 {
```
aleksej-paschenko, line 70: how about starting the function with this check?

cmd/ketch/app_create.go
```diff
@@ -54,6 +57,11 @@ func appCreate(ctx context.Context, cfg config, options appCreateOptions, out io
 	if err != nil {
 		return fmt.Errorf("failed to parse env variables: %w", err)
 	}
+	if options.platform != "" {
```
aleksej-paschenko, line 60: looks like this is from another commit

aleksej-paschenko (changes requested):

cmd/ketch/app_list.go
```diff
@@ -42,14 +44,61 @@ func appList(ctx context.Context, cfg config, out io.Writer) error {
...
+		return nil, nil
+	}
+	selector := &metav1.LabelSelector{
+		MatchExpressions: []metav1.LabelSelectorRequirement{
+			{
+				Key:      ketchAppNameLabel,
+				Operator: "In",
```
aleksej-paschenko, line 77: there is `Exists` operator and you can use it without the app names list

aleksej-paschenko (changes requested):

cmd/ketch/app.go
c12o16h1: Well, in this case, we will use "apiv1" statuses but will mutate the original state. So the state in Ketch will be different from the state in Kubernetes while naming be the same. So, this will confuse the user (it will confuse even me with time), because you need to know that the state in ketch is different from the state in Kubernetes. Sure, we may assume that users know all ketch documentation, but they will not, and we don't have such good documentation, yet. 

Separate naming is better in this case, IMHO.

aleksej-paschenko (changes requested):

cmd/ketch/app_list.go
```diff
@@ -42,14 +44,57 @@ func appList(ctx context.Context, cfg config, out io.Writer) error {
...
 	}
 	w.Flush()
 	return nil
 }
+
+func allAppsPods(ctx context.Context, cfg config, apps []ketchv1.App) (*corev1.PodList, error) {
+	if len(apps) == 0 {
+		return nil, nil
+	}
+	var appNames []string
```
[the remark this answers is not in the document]
c12o16h1: indeed, forget to clean that part. thanks.

cmd/ketch/app_list.go
```diff
@@ -42,14 +44,57 @@ func appList(ctx context.Context, cfg config, out io.Writer) error {
...
+		LabelSelector: s.String(),
+	})
+}
+
+func filterAppPods(appName string, pods []corev1.Pod) []corev1.Pod {
+	var appPods []corev1.Pod
+	for _, pod := range pods {
+		if pod.Labels[ketchAppNameLabel] == appName {
+			appPods = append(appPods, pod)
+			continue
```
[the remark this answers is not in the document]
c12o16h1: correct, thanks

cmd/ketch/app_list.go
```diff
@@ -42,14 +44,52 @@ func appList(ctx context.Context, cfg config, out io.Writer) error {
...
 	}
+	allPods, err := allAppsPods(ctx, cfg, apps.Items)
+	if err != nil {
+		return fmt.Errorf("failed to list apps pods: %w", err)
+	}
 	w := tabwriter.NewWriter(out, 0, 4, 4, ' ', 0)
-	fmt.Fprintln(w, "NAME\tPOOL\tUNITS\tADDRESSES\tPLATFORM\tDESCRIPTION")
+	fmt.Fprintln(w, "NAME\tPOOL\tSTATE\tADDRESSES\tPLATFORM\tDESCRIPTION")
 	for _, item := range apps.Items {
+		pods := filterAppPods(item.Name, allPods.Items)
```
aleksej-paschenko, line 54: allPods can be nil here

aleksej-paschenko (changes requested):

cmd/ketch/app.go
```diff
@@ -26,3 +31,80 @@ func newAppCmd(cfg config, out io.Writer) *cobra.Command {
...
+	}
+	if succeeded > 0 {
+		parts = append(parts, fmt.Sprintf(`%d %s`, succeeded, ketchv1.PodSucceeded))
+	}
+
+	return strings.Join(parts, ", ")
+}
+
+var (
+	containerStateCrashLoopBackOff = "CrashLoopBackOff"
```
murphybytes, line 76: These look like they should be constants

murphybytes (approved):

aleksej-paschenko (approved):

PR #50, opened 2021-01-21
[SHIPA-720] Deploy from source.

murphybytes:
# Description

See [SHIPA-720](https://shipaio.atlassian.net/browse/SHIPA-720?atlOrigin=eyJpIjoiMzVkNjJjMGExZmQyNGVkMmEzOWRmOGVkZWViNjA3MTQiLCJwIjoiaiJ9).  This PR allows user to specify a source directory containing code that will be built and deployed for an app.

## Type of change

## Testing

## Documentation

## Final Checklist:


internal/archive/tar.go
```diff
@@ -0,0 +1,185 @@
...
+	for _, input := range inputs {
+		input(&options)
+	}
+	// if no directories or files are specified, default to current directory
+	if len(options.dirs) == 0 && len(options.files) == 0 {
+		options.dirs = []string{"."}
+	}
+
+	// change to working directory
+	currentWd, err := os.Getwd()
```
itsksaurabh, line 97: not sure if you need to call `os.Getwd()` multiple times here if [options.workingDir](https://github.com/shipa-corp/ketch/pull/50/files#diff-f0d23faf1e41e68a9840cedd68384283a0fd8beb383eb40299fa5b21e7be604eR85) gets the same value.

internal/build/build.go
```diff
@@ -0,0 +1,224 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	ketchv1 "github.com/shipa-corp/ketch/internal/api/v1beta1"
+	"github.com/shipa-corp/ketch/internal/archive"
+	"github.com/shipa-corp/ketch/internal/docker"
+	"github.com/shipa-corp/ketch/internal/errors"
+)
+
+const (
+	archiveFileName      = "archive.tar.gz"
```
itsksaurabh, line 23: We define it [here](https://github.com/shipa-corp/ketch/pull/50/files#diff-ebba829eab1ca7dd99d66964c8eddd4febfda8f9dd35dbcbd28e78cece01d37bR15). Is it possible to define such constants once and reuse them?

internal/archive/tar.go
```diff
@@ -0,0 +1,185 @@
...
+	var err error
+	if options.workingDir, err = os.Getwd(); err != nil {
+		return errors.Wrap(err, "get working dir failed")
+	}
+	for _, input := range inputs {
+		input(&options)
+	}
+	// if no directories or files are specified, default to current directory
+	if len(options.dirs) == 0 && len(options.files) == 0 {
+		options.dirs = []string{"."}
```
itsksaurabh, line 93: how about validating path strings provided by the users?

internal/build/build.go
```diff
@@ -0,0 +1,224 @@
...
+	out io.Writer
+}
+
+// CreateImageFromSourceResponse is returned from the build handler function and contains the
+// fully qualified image name that was built.
+type CreateImageFromSourceResponse struct {
+	ImageURI string
+}
+
+type Option func(o *CreateImageFromSourceRequest)
```
itsksaurabh, line 59: since it's exported, we should add comments on it.

internal/archive/tar.go
```diff
@@ -0,0 +1,185 @@
...
+// rooted at the current directory. withWorkingDirectory is used to set working directory and defaults
+// to the current working directory. The archival process can be instructed to exclude certain files by placing a
+// special file .shipaignore in the working directory.  It uses pattern matches like .gitignore such the files
+// matching the patterns are excluded from the archive. Files specified with includeFiles override .shipaignore and
+// are added to the archive.
+func Create(archiveFile string, inputs ...Option) error {
+	var options tarOptions
+	var err error
+	if options.workingDir, err = os.Getwd(); err != nil {
+		return errors.Wrap(err, "get working dir failed")
```
itsksaurabh, line 86: "could not get working directory" ?

internal/build/build.go
```diff
@@ -0,0 +1,224 @@
...
+		if err != nil {
+			return nil, errors.Wrap(err, "could not get working directory")
+		}
+
+		req.workingDir = wd
+		// build output default to stdout
+		req.out = os.Stdout
+
+		// source code default directory is current dir
+		req.sourcePaths = []string{"."}
```
itsksaurabh, line 98: should we make default values constants?

internal/build/build.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// BuildDir contains directory where generated Dockerfile and source archive is located
+func (bc *buildContext) BuildDir() string {
+	return bc.ephemeralBuildDir
+}
+
+// Prepare the build. Create a temp directory containing a docker file, and an archive containing source codes.
+func (bc *buildContext) prepare(platformImage string, workingDir string, sourcePaths []string) error {
+	const sourceDockerfileTemplate = `FROM {{ .PlatformImage }}
```
itsksaurabh, line 183: I think we should avoid using hardcoded string configs.

itsksaurabh, line 183: I think we should avoid using hardcoded string configs.

itsksaurabh (changes requested):

internal/build/hooks.go
```diff
@@ -0,0 +1,60 @@
...
+	"io/ioutil"
+	"os"
+	"path"
+	"sigs.k8s.io/yaml"
+
+	"github.com/shipa-corp/ketch/internal/errors"
+)
+
+// Shipa holds the decoded results of a shipa.yaml file
+type Shipa struct {
```
aleksej-paschenko, line 13: there is already a type for this https://github.com/shipa-corp/ketch/blob/main/internal/api/v1beta1/ketch_yaml.go#L17
I would say that we should use it and add `Build []string` to the type.

```bash
ketch app deploy --ketch-yaml=<path-to-ketch-yaml> --strict
```
With --strict parameter set ketch validates `ketch.yaml` and it would be great to validate hooks too.

internal/build/hooks.go
```diff
@@ -0,0 +1,60 @@
...
+	Hooks *Hooks `json:"hooks,omitempty"`
+}
+
+// Hooks contain build hooks from a shipa yaml file. Hooks are shell scripts that
+// run during deployments. See https://learn.shipa.io/docs/shipayml
```
aleksej-paschenko, line 18: can we use https://learn.theketch.io/docs/ketchyaml?

internal/build/hooks.go
```diff
@@ -0,0 +1,60 @@
...
+type Hooks struct {
+	Build []string `json:"build,omitempty"`
+}
+
+func getHooks(workingDir string, shipaFileSearchPaths []string) ([]string, error) {
```
aleksej-paschenko, line 23: I think that we can validate and decode `ketch.yaml` at the begging of `appDeploy` function and then pass it to `GetSourceHandler` and `appDeployImage`

internal/archive/tar.go
```diff
@@ -0,0 +1,185 @@
...
+	ignore "github.com/sabhiram/go-gitignore"
+	"io"
+	"os"
+	"path/filepath"
+
+	"github.com/shipa-corp/ketch/internal/errors"
+)
+
+const (
+	shipaIgnoreFile = ".shipaignore"
```
aleksej-paschenko, line 18: ketch is a standalone product. For example, it uses `ketch.yaml` instead of `shipa.yaml`. 
Not sure about `.ketchignore`, it might be that` []string{'.shipaignore', '.ketchignore'}` should be used. Let's discuss it

aleksej-paschenko (changes requested):
looks awesome!

cmd/ketch/app_deploy.go
```diff
@@ -16,30 +16,46 @@ import (
...
 	"github.com/shipa-corp/ketch/internal/chart"
+	"github.com/shipa-corp/ketch/internal/docker"
 )
 
 const appDeployHelp = `
-Roll out a new version of an application with an image. 
+Roll out a new version of an application with an image.
+
+Deploy from source code, <source> is path to source code or a zip formatted file :
+  ketch app deploy <app name> <source> -i myregistry/myimage:latest
```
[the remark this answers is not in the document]
aleksej-paschenko: that's what confused me too, then I realized that if I provide an image and source path, ketch builds a new image from sources and pushes it to the provided image. So we are going to use the same flag for different things `--image=<source-image>` and `--image=<dest-image>`

internal/build/build.go
murphybytes: I disagree.  This string is small, and it is only used within the scope of the method where it is called.  I was going to use gobuffalo/packr to embed this in the binary, but I found out that go:embed will be included in Go 1.16, which is much easier to use and will be a feature of the language moving forward. So in short, leave as is until Go 1.16 drops.  I put a todo comment in the code indicating as much.

aleksej-paschenko (approved):

PR #60, opened 2021-02-01
[SHIPA-1016] fix 404 error for Canary deployment 

itsksaurabh:
# Description

This PR fixes the issue that during Canary deployment we were getting 404 error.

According to my observation, It was due to the following:

We have two ways of changing an app CRD:
1. `app-deploy` command
2. `App controller`

## issues

- With `app deploy` command we were changing the traffic weight for the canary deployment and we were assigning the canary weight without making sure that the deployment pods are in a running state. Hence it was causing issue as the redirection was to the pod which was still building itself.

- Once the canary deployment is initiated, we let the `App controller` change the CRD, It also kept changing the weight periodically even when the canary deployment is at non-running state. Since, we can't stop the canary as of now, it may bring the app to a stage when 100% traffic will go to canary deployment which might be in a non-running state.

- For Traefik ingressRoute, if weights are distributed across services and the pods for deployment for the canary service are not running then it gives 404 error.

So even if you set the canary deployment weight to `0` and pods for it is in a non-running state, it will give 404 until the pods come to a running state.

## Fixes
- `app deploy` doesn't change the routing weight initially. So we set initial weight for canary deployment to zero. App controller will update the weight once all pods for canary will be on running state.

- For `App controller` we first assure that all pods for the canary deployment are in running state, then we change update the traffic weight.

- In Traefik and Istio CRDs ( IngressRoute and Virtual Services ), we do not add a service or route until the weight of the traffic is greater than 0. In this way, we only add them when the services require some traffic. 

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/internal/controllers/app_controller.go
+++ b/internal/controllers/app_controller.go
@@ -32,7 +32,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/tools/reference"
-	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -40,6 +39,9 @@ import (
 	"github.com/shipa-corp/ketch/internal/chart"
 	"github.com/shipa-corp/ketch/internal/templates"
 )
+
+// reconcileTimeout is the timeout to trigger Operator reconcile
+const reconcileTimeout = 30 * time.Second
 
 // AppReconciler reconciles a App object.
 type AppReconciler struct {
@@ -121,7 +123,7 @@ func (r *AppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 
 	if app.Spec.Canary.Active {
-		result = ctrl.Result{RequeueAfter: app.Spec.Canary.NextScheduledTime.Sub(r.Now())}
+		result = ctrl.Result{RequeueAfter: reconcileTimeout}
 	}
 
 	return result, err
@@ -202,14 +204,7 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
 	// check for canary deployment
 	if app.Spec.Canary.Active {
 		// retry until all pods for canary deployment comes to running state.
-		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if err := canaryPodStatus(r, app); err != nil {
-				return chart.ErrEmptyProcfile
-			}
-			return nil
-		})
-
-		if retryErr != nil {
+		if err := canaryPodStatus(r, app); err != nil {
 			return reconcileResult{
 				status:  v1.ConditionFalse,
 				message: fmt.Sprintf("canary upgrade failed: %v", err),
--- a/internal/controllers/app_controller.go
+++ b/internal/controllers/app_controller.go
@@ -244,13 +244,16 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
 	}
 }
 
+// canaryPodStatus checks whether all pods for canary deployment are running or not.
 func canaryPodStatus(r *AppReconciler, app *ketchv1.App) error {
 	if len(app.Spec.Deployments) <= 1 {
 		return errors.New("no canary deployment found")
 	}
 
+	// podList contains list of Pods matching the specifed labels below
 	podList := &v1.PodList{}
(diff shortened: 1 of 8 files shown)


internal/controllers/app_controller.go
```diff
@@ -200,6 +201,22 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
 
 	// check for canary deployment
 	if app.Spec.Canary.Active {
+		// retry until all pods for canary deployment comes to running state.
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			if err := canaryPodStatus(r, app); err != nil {
+				return chart.ErrEmptyProcfile
```
aleksej-paschenko, line 207: ErrEmptyProcfile?

internal/controllers/app_controller.go
```diff
@@ -200,6 +201,22 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
 
 	// check for canary deployment
 	if app.Spec.Canary.Active {
+		// retry until all pods for canary deployment comes to running state.
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
aleksej-paschenko, line 205: `canaryPodStatus` doesn't update anything, let's use a loop there

aleksej-paschenko (changes requested):

internal/controllers/app_controller.go
itsksaurabh: ah , by mistake I put it there. Fixing it. Sorry.

internal/controllers/app_controller.go
```diff
@@ -227,6 +241,34 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
 	}
 }
 
+// canaryPodStatus checks whether all pods for canary deployment are running or not.
+func canaryPodStatus(r *AppReconciler, app *ketchv1.App) error {
```
aleksej-paschenko, line 245: can we pass `List` function there or maybe the client

internal/controllers/app_controller.go
```diff
@@ -119,17 +122,18 @@ func (r *AppReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return result, err
 	}
 
-	if app.Spec.Canary.Active {
-		result = ctrl.Result{RequeueAfter: app.Spec.Canary.NextScheduledTime.Sub(r.Now())}
+	if scheduleResult.useTimeout {
```
aleksej-paschenko, line 125: what about canary.Active? what about step intervals?

aleksej-paschenko (changes requested):

internal/controllers/app_controller.go
```diff
@@ -227,6 +247,34 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
...
+	}
+
+	if err := c.List(context.Background(), podList, listOpts...); err != nil {
+		return err
+	}
+
+	// check if all pods are running for the deployment
+	for _, pod := range podList.Items {
+		if pod.Status.Phase != v1.PodRunning {
+			return errors.New("all pods are not running")
```
c12o16h1, line 272: What about moving this error, and error at the begin of function into variables?

internal/controllers/app_controller.go
```diff
@@ -200,6 +210,16 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
 
 	// check for canary deployment
 	if app.Spec.Canary.Active {
+		// retry until all pods for canary deployment comes to running state.
+		if err := canaryPodStatus(r.Client, app); err != nil {
+			return reconcileResult{
+				status:     v1.ConditionFalse,
+				message:    fmt.Sprintf("canary upgrade failed: %v", err),
```
c12o16h1, line 217: Why `upgrade`? Do we upgade anything here? Sounds like we tried to upgrade canary itself, but failed.

c12o16h1 (changes requested):
texts

internal/controllers/app_controller.go
itsksaurabh: Since we are not reusing this error now, so it's ok to keep it here. :)

internal/controllers/app_controller.go
itsksaurabh: I would like to clarify that we update the canary weights after step interval. I agree with you, I would use `update` instead. Thanks!

aleksej-paschenko (approved):

c12o16h1 (approved):

PR #63, opened 2021-02-03
[Shipa-1026]

itsksaurabh:
# Description

Fixes  https://shipaio.atlassian.net/browse/SHIPA-1026

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/internal/api/v1beta1/app_types.go
+++ b/internal/api/v1beta1/app_types.go
@@ -34,6 +34,9 @@ func init() {
 const (
 	ShipaCloudDomain     = "shipa.cloud"
 	DefaultNumberOfUnits = 1
+	// CanaryFailureCountLimit is the maximum number of failure Count for canary deployment.
+	// If it reaches this limit then weights will be rolledback to primary deployment and canary will become inactive
+	CanaryFailureCountLimit = 5
 )
 
 // Env represents an environment variable present in an application.
@@ -183,6 +186,10 @@ type CanarySpec struct {
 	CurrentStep int `json:"currentStep,omitempty"`
 	// Active shows if canary deployment is active for this application.
 	Active bool `json:"active,omitempty"`
+	// FailureCount is a number that holds how many times the canary deployment was in unhealthy state.
+	// This will help to rollback all the waits to primary deployment if the count reaches it's limit.
+	// +kubebuilder:validation:Minimum=0
+	FailureCount int `json:"failureCount,omitempty"`
 }
 
 // AppSpec defines the desired state of App.
--- a/internal/api/v1beta1/app_types.go
+++ b/internal/api/v1beta1/app_types.go
@@ -190,6 +190,8 @@ type CanarySpec struct {
 	// This will help to rollback all the waits to primary deployment if the count reaches it's limit.
 	// +kubebuilder:validation:Minimum=0
 	FailureCount int `json:"failureCount,omitempty"`
+	// Timeout for canary deployment
+	Timeout time.Duration `json:"timeout,omitempty"`
 }
 
 // AppSpec defines the desired state of App.
--- a/internal/api/v1beta1/app_types.go
+++ b/internal/api/v1beta1/app_types.go
@@ -575,9 +575,9 @@ func (app *App) DoCanary(now metav1.Time) error {
 		}
 	}
 
-	// update failure count on successful canary rollout
+	// reset failure count on successful canary rollout
 	if app.Spec.Canary.FailureCount > 0 {
-		app.Spec.Canary.FailureCount--
+		app.Spec.Canary.FailureCount = 0
 	}
 
 	return nil
--- a/internal/api/v1beta1/app_types.go
+++ b/internal/api/v1beta1/app_types.go
@@ -36,7 +36,7 @@ const (
 	DefaultNumberOfUnits = 1
 	// CanaryFailureCountLimit is the maximum number of consecutive failure Count for canary deployment.
 	// If it reaches this limit then weights will be rolledback to primary deployment and canary will become inactive
-	CanaryFailureCountLimit = 20
+	CanaryFailureCountLimit = 30
 )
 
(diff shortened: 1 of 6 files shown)


internal/controllers/app_controller.go
```diff
@@ -210,8 +207,27 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
...
+				status:     v1.ConditionFalse,
+				message:    "no canary deployment found",
+				useTimeout: true,
+			}
+		}
+
 		// retry until all pods for canary deployment comes to running state.
 		if err := checkPodStatus(r.Client, app.Name, app.Spec.Deployments[1].Version); err != nil {
+			// check for rollback
+			app.CheckForRollback()
```
aleksej-paschenko, line 223: it doesn't `check` for rollback it does it or increments the failure counter. Is it possible to have a more readable approach like
```golang
if err := checkPodStatus(r.Client, app.Name, app.Spec.Deployments[1].Version); err != nil {
    app.IncrementCanaryFailureCounter()
    app.DoRollbackIfNeeded()
}

```

internal/controllers/app_controller.go
```diff
@@ -210,8 +207,27 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
...
+			if e := r.Update(ctx, app); err != nil {
+				return reconcileResult{
+					status:  v1.ConditionFalse,
+					message: fmt.Sprintf("failed to perform rollback: %v", e),
```
aleksej-paschenko, line 227: because `CheckForRollback` function implements two different things this error message has to be "failed to perform rollback" or "failed to increment canary failure counter"

internal/api/v1beta1/app_types.go
```diff
@@ -565,9 +574,29 @@ func (app *App) DoCanary(now metav1.Time) error {
 			app.Spec.Deployments = []AppDeploymentSpec{app.Spec.Deployments[1]}
 		}
 	}
+
+	// update failure count on successful canary rollout
+	if app.Spec.Canary.FailureCount > 0 {
```
aleksej-paschenko, line 579: why we should do it?

internal/controllers/app_controller.go
```diff
@@ -275,6 +291,12 @@ func checkPodStatus(c client.Client, appName string, depVersion ketchv1.Deployme
 		if pod.Status.Phase != v1.PodRunning {
 			return errors.New("all pods are not running")
 		}
+
+		for _, c := range pod.Status.Conditions {
```
aleksej-paschenko, line 295: what if an application consists of several containers and one of them has a script that performs an action and exits.

aleksej-paschenko (changes requested):

internal/api/v1beta1/app_types.go
itsksaurabh: I was thinking about consecutive failures, so idea is to reset it on successful rollouts

internal/api/v1beta1/app_types.go
```diff
@@ -565,9 +572,32 @@ func (app *App) DoCanary(now metav1.Time) error {
 			app.Spec.Deployments = []AppDeploymentSpec{app.Spec.Deployments[1]}
 		}
 	}
+
+	// reset failure count on successful canary rollout
+	if app.Spec.Canary.FailureCount > 0 {
```
aleksej-paschenko, line 577: do we need `FailureCount > 0` check?

internal/controllers/app_controller.go
```diff
@@ -210,8 +207,38 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
 
 	// check for canary deployment
 	if app.Spec.Canary.Active {
+		// ensures that the canary deployment exists
+		if len(app.Spec.Deployments) <= 1 {
+			app.Spec.Canary.Active = false
```
aleksej-paschenko, line 212: only app.Status is updated if this function returns an error

internal/controllers/app_controller.go
```diff
@@ -210,8 +207,38 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App) reconci
...
+			if e := r.Update(ctx, app); err != nil {
+				return reconcileResult{
+					status:     v1.ConditionFalse,
+					message:    fmt.Sprintf("failed to increment canary failure count: %v", e),
+					useTimeout: true,
+				}
+			}
+
+			// check if rollback is needed
+			app.DoRollbackIfNeeded()
```
aleksej-paschenko, line 233: can we have only one Update operation or update the second time only if we have to update the app. 

```golang

app.IncrementFailureCount()
app.DoRollbackIfNeeded() 
r.update()..
```

internal/controllers/app_controller.go
```diff
@@ -272,9 +299,20 @@ func checkPodStatus(c client.Client, appName string, depVersion ketchv1.Deployme
 
 	// check if all pods are running for the deployment
 	for _, pod := range podList.Items {
+		// check if pod have voluntarily terminated with a container exit code of 0
+		if pod.Status.Phase == v1.PodSucceeded {
```
aleksej-paschenko, line 303: not sure that we have good coverage for ready conditions. 
if "web" process has PodSucceeded status then how will it respond to user requests?
if "worker" process has PodSucceeded status then is it ok to return nil? "web" process could not be ready yet

aleksej-paschenko (changes requested):

internal/controllers/app_controller.go
itsksaurabh: `PodSucceeded ` is the phase that indicates if the pod has voluntarily terminated with a container exit code of 0. If it is terminated with some other error code then this condition will not match. The above examples seem more like a user-side error.

aleksej-paschenko (approved):

PR #94, opened 2021-06-10
Integration tests

stinkyfingers:
# Description

Adds Ketch CLI Integration tests.

Fixes # (issue)

## Type of change

- [x] Chore (documentation addition or typo, file relocation)

## Testing

## Documentation

## Final Checklist:


cli_tests/app.sh
```diff
@@ -0,0 +1,172 @@
...
+
+# To run locally:
+# export KETCH_COMMAND=<location of ketch binary>
+# assure you have a kubernetes cluster running w/ traefik, cert manager, etc. (see ketch getting started docs)
+# assure the ketch cli is compiled (make ketch)
+# assure you have bats installed locally (via apt, brew, etc.)
+# ./cli_tests/app.sh
+
+setup() {
+    if [[ -z "${KETCH_COMMAND}" ]]; then
```
DavisFrench, line 11: I feel like `KETCH_COMMAND` is a little misleading since I thought it was referring to something like app or framework at first. Maybe a variable name like KETCH_EXECUTABLE_PATH or something would be a bit more clear?

cli_tests/app.sh
```diff
@@ -0,0 +1,172 @@
...
+
+@test "framework update" {
+  result=$($KETCH framework update "$FRAMEWORK" --app-quota-limit 1)
+  echo "RECEIVED:" $result
+  [[ $result =~ "Successfully updated!" ]]
+}
+
+@test "app deploy" {
+  run $KETCH app deploy "$APP_NAME" --framework "$FRAMEWORK" -i "$APP_IMAGE"
+  [ $status -eq 0 ]
```
DavisFrench, line 56: I'm a bit curious, but why are you making use of single brackets here as opposed to double? I understand there are some portability issue when making use of double brackets, but the regex operator is making use of double square brackets, so I'm wondering if there is some nuance I'm missing?

cli_tests/app.sh
```diff
@@ -0,0 +1,172 @@
...
+  echo "RECEIVED:" $result
+  [[ ! $result =~ "Address: http://$CNAME" ]]
+}
+
+@test "unit add" {
+ run $KETCH unit add 1 --app "$APP_NAME"
+ [ $status -eq 0 ]
+ result=$(kubectl describe apps $APP_NAME)
+ echo "RECEIVED:" $result
+ [[ $result =~ "Units:  2" ]] # note two spaces
```
DavisFrench, line 124: Thank you for these comments, the double spaces look like a huge source of frustration for debugging if you don't know about them.

cli_tests/install_ketch.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# kustomize
+make install-kustomize KUSTOMIZE_INSTALL_DIR=/tmp
+export PATH=$PATH:/tmp
+
+# ketch
+kubectl cluster-info
+make manifests install ketch manager
+export PATH=$PATH:$(pwd)/bin
+ketch -v
```
DavisFrench, line 25: This version check is just to confirm if everything is copacetic with ketch correct (set -e makes it so that if this command fails it exits)?

cli_tests/install_kubectl.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.20.1/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
```
DavisFrench, line 3: This seems like a pretty long series of commands for just one line. Doing something like
```
curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.20.1/bin/linux/amd64/kubectl && \
chmod +x kubectl && \
sudo mv kubectl /usr/local/bin/
```
might make it a bit easier to read.

cli_tests/install_minikube.sh
```diff
@@ -0,0 +1,15 @@
...
+
+export CHANGE_MINIKUBE_NONE_USER=true
+export MINIKUBE_WANTUPDATENOTIFICATION=false
+export MINIKUBE_WANTREPORTERRORPROMPT=false
+export MINIKUBE_HOME=$HOME
+export KUBECONFIG=$HOME/.kube/config
+
+# Minikube
+sudo apt-get update -yq && sudo apt-get -qq -y install conntrack
+curl -Lo minikube https://storage.googleapis.com/minikube/releases/v1.20.0/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
```
DavisFrench, line 13: I think this might also be a bit too long for one line.

DavisFrench (approved):
Everything looks good! All of my comments boil down to minor nit picks or just something that I want to know for my own benefit.

cli_tests/app.sh
```diff
@@ -0,0 +1,172 @@
...
+
+setup() {
+    if [[ -z "${KETCH_COMMAND}" ]]; then
+    KETCH=$(pwd)/bin/ketch
+  else
+    KETCH="${KETCH_COMMAND}"
+  fi
+  INGRESS=$(kubectl get svc traefik -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
+  FRAMEWORK="myframework"
+  APP_IMAGE="docker.io/shipasoftware/bulletinboard:1.0"
```
itsksaurabh, line 18: could you build a lite-weight image for testing purposes instead of this, especially for tests? Runners usually have very low CPU and memory, any such improvement would be nice! :)

cli_tests/app.sh
itsksaurabh, line 56: can we check the final results output here? Something like "Successfully deployed!" or whatever ketch prints?

cli_tests/app.sh
```diff
@@ -0,0 +1,172 @@
...
+
+@test "app start" {
+  result=$($KETCH app start "$APP_NAME")
+  echo "RECEIVED:" $result
+  [[ $result =~ "Successfully started!" ]]
+}
+
+@test "app log" {
+  run $KETCH app log "$APP_NAME"
+  [ $status -eq 0 ]
```
itsksaurabh, line 91: I am not sure how does it work. where did we define this variable?

I generally have seen it is set using:
```
## get status ##
status=$?
```
or

```
#!/bin/bash
/path/to/some/command
[ $? -eq 0 ]  || exit 1
```

itsksaurabh: what will happen if `[ $status -eq 1 ]`? Will it fail?

cli_tests/install_ketch.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+# exit when any command fails
+set -e
```
itsksaurabh, line 4: how about also adding `set -o pipefail`? This setting prevents errors in a pipeline from being masked. If any command in a pipeline fails, that return code will be used as the return code of the whole pipeline. By default, the pipeline's return code is that of the last command even if it succeeds.

itsksaurabh: Also please consider `set -x`. It enables a mode of the shell where all executed commands are printed to the terminal. In our case, it would be helpful for debugging, which is a typical use case for set -x printing every command as it is executed may help us to visualize the control flow of the script if it is not functioning as expected.

cli_tests/app.sh
stinkyfingers: I like it. Change made.

cli_tests/app.sh
stinkyfingers: The double brackets were required for the regex comparisons: `=~`. I made the `status` equality checks double too, for consistency. Note: I'm not a bash expert.

cli_tests/install_ketch.sh
stinkyfingers: Yeah - double checks on the off chance the install steps go fine but `ketch` is still missing for some reason.

cli_tests/install_kubectl.sh
stinkyfingers: Good call. Legible-ified.

cli_tests/install_minikube.sh
stinkyfingers: Change made.

cli_tests/app.sh
stinkyfingers: I sure can. I pushed a super-quick image to our gcp CI GCR repo for testing. I'm hoping that is what you're after.

cli_tests/app.sh
stinkyfingers: I only checked the status code here, and for a few other commands, because ketch does not seem to print anything on success. Also, the `app list` test will assert that the app was actually created.

cli_tests/app.sh
stinkyfingers: `$status` is the command's exits status. It gets populated by the `BATS` testing framework when using the `run` command specifically for this type of check. `[ $status -eq 1 ]` will fail. I'm _pretty_ sure this is the author's intent: https://bats-core.readthedocs.io/en/stable/writing-tests.html#run-test-other-commands

cli_tests/install_ketch.sh
stinkyfingers: Good call. Added `x` and `o`.

PR #95, opened 2021-06-10
Refactors CLI output into extensible package

stinkyfingers:
# Description

We discussed adding an `--output json/yaml` flag to the CLI, similar to kubectl, but decided to hold off as we are about to implement exporting of yaml configs for App and Framework and this could be counterintuitive to users. This PR refactors CLI output into a separate extensible package to help clean up code and in case we opt to add output formatting later. Adds a custom `column` marshaler. 

## Type of change

## Testing

## Documentation

## Final Checklist:


DavisFrench (commented):
Outside of the one ignored error, looks great! The rest of my comments boil down to nit picks. Really excited for this feature to be merged in!

cmd/ketch/root.go
```diff
@@ -57,6 +57,8 @@ func newRootCmd(cfg config, out io.Writer, packSvc *pack.Client, ketchConfig con
 			return cmd.Usage()
 		},
 	}
+
+	cmd.PersistentFlags().StringP("output", "o", "", "used to specify output, e.g. --output format=json")
```
DavisFrench, line 61: Does this technically count as a global command? From the documentation, this will be available on all commands, but should this be expected to work on something like `ketch framework add`?

cmd/ketch/app_info.go
```diff
@@ -92,46 +103,56 @@ func appInfo(ctx context.Context, cfg config, options appInfoOptions, out io.Wri
...
 	appPods, err := cfg.KubernetesClient().CoreV1().Pods(app.Namespace).List(ctx, metav1.ListOptions{
 		LabelSelector: fmt.Sprintf(`%s=%s`, utils.KetchAppNameLabel, app.Name),
 	})
 	if err != nil {
 		return err
 	}
-	fmt.Fprintln(w, "DEPLOYMENT VERSION\tIMAGE\tPROCESS NAME\tWEIGHT\tSTATE\tCMD")
+
+	data := generateAppInfoOutput(app, appPods, framework)
+	outputFlag, _ := flags.GetString("output")
```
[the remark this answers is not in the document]
stinkyfingers: Whoops, good catch. Let's actually check the error.

cmd/ketch/env_get.go
```diff
@@ -5,6 +5,10 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/shipa-corp/ketch/cmd/ketch/output"
```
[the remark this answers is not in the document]
stinkyfingers: Nitpick away. Fixed. Goimport was dumping them in a new block.

cmd/ketch/root.go
stinkyfingers: This is a good call. I went back & forth on whether to add output formatting to every command (`{"status":"Successfully added framework!"}`?) I moved the flag to the affected commands.

DavisFrench (approved):

cmd/ketch/app_info.go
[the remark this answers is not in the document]
aleksej-paschenko: not sure that I get it, so what's the difference between
```golang
	outputFlag, err := flags.GetString("output")
	if err != nil {
		outputFlag = ""
	}
        // vs
	outputFlag, _ := flags.GetString("output")


```

cmd/ketch/app_info.go
```diff
@@ -59,6 +57,20 @@ type appInfoContext struct {
 	Table       string
```
aleksej-paschenko, line 57: `Table` is not used anymore

cmd/ketch/app_info.go
```diff
@@ -59,6 +57,20 @@ type appInfoContext struct {
...
 }
 
+type appInfoOutput struct {
+	AppInfoContext appInfoContext
```
aleksej-paschenko, line 61: how about having yaml and json tags for `AppInfoContext` and `Deployments`?

cmd/ketch/output/output.go
```diff
@@ -0,0 +1,37 @@
...
+	"io"
+
+	"github.com/spf13/pflag"
+)
+
+type Writer interface {
+	Write() error
+}
+
+func Write(data interface{}, out io.Writer, flags *pflag.FlagSet) error {
```
aleksej-paschenko, line 13: can this function take in `outputFlag string`? 
It's a utility function and there is no need to share information about the flagset.

cmd/ketch/output/yaml.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	"fmt"
+	"io"
+
+	"sigs.k8s.io/yaml"
+)
+
+// YAML represents data and a writer for yaml output type
+type YAML struct {
```
aleksej-paschenko, line 11: is there a chance to start with unexported types? 
`YAML`, `JSON` and `Column` are consumed by this package only

```
type yaml struct {
	Data   interface{}
	Writer io.Writer
}

func (y *yaml) Write() error {
	d, err := yaml.Marshal(y.Data)
	if err != nil {
		return err
	}
	_, err = fmt.Fprintln(y.Writer, string(d))
	return err
}
```

cmd/ketch/output/output.go
```diff
@@ -0,0 +1,37 @@
+package output
+
+import (
+	"io"
+
+	"github.com/spf13/pflag"
+)
+
+type Writer interface {
```
aleksej-paschenko, line 9: I don't think we need to export this interface, it's used only by the `Write` function.
it can be 
```golang
	type writer interface {
		Write() error
	}
	var w writer

```

aleksej-paschenko (changes requested):
The Column thing rocks)
I've left several comments to adderss but my main question is what are the use-cases for --output?

Given that we are going to introduce `app.yaml` and `framework.yaml`, it would be great to see how both features play together.

cmd/ketch/app_info.go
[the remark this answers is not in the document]
stinkyfingers: I switched to the latter. I'm always torn between making code _really_ explicit and simplifying it.

cmd/ketch/app_info.go
stinkyfingers: removed

cmd/ketch/app_info.go
stinkyfingers: Added. I added struct tags to builder and env types too.

cmd/ketch/output/output.go
stinkyfingers: Good call. Change made.

cmd/ketch/output/yaml.go
stinkyfingers: Good call. I privatized everything except the one `Write()`.

cmd/ketch/output/output.go
stinkyfingers: Change made.

stinkyfingers:
@aleksej-paschenko I don't have a GREAT use case. At some point I was scripting something and wanted output like `ketch framework list -o json | jq '.[0].namespace'` (I don't remember what I was doing specifically). It seemed fairly easy to implement and intuitive for people familiar with `kubectl`.

aleksej-paschenko (approved):

PR #96, opened 2021-06-11
Cli shell autocomplete

koncar:
# Description

Implement shell autocomplete

## Type of change

## Testing

## Documentation

## Final Checklist:


cmd/ketch/autocompletion.go
```diff
@@ -0,0 +1,89 @@
+package main
```
DavisFrench, line 1: This is more of a nit pick than an actual issue, but I think auto_completion.go may be a better name for the file. I think it would be more in line with the naming convention already present in the repo.

cmd/ketch/builder_list_test.go
```diff
@@ -67,3 +68,52 @@ func TestBuilderList(t *testing.T) {
...
+			builderList: BuilderList{
+				{Image: "img1"},
+				{Image: "img2"},
+				{Image: "img3"},
+			},
+			filter: nil,
+			want:   []string{"img1", "img2", "img3"},
+		},
+		{
+			name: "filtered result",
```
DavisFrench, line 91: This test seems to be the same as the one above it. Based off the name, I think that the filter should maybe be something like "img1" instead of nil.

internal/deploy/parameters.go
```diff
@@ -22,6 +22,7 @@ import (
 
 // Command line flags
 const (
+	FlagApp            = "app"
```
DavisFrench, line 25: I like the reuse of this variable across multiple files, but I think that this is a bit out of the scope for this file. Do you think that in the future moving all reused flags to a distinct package would be a good idea?

DavisFrench: Originally the app_deploy code got moved into the deploy directory, but through suggestions on the pr it got moved back to ketch/. These flags were originally private to deploy, but had to be made public because of this moving of the functionality back to ketch/.

cmd/ketch/autocompletion_test.go
```diff
@@ -0,0 +1,103 @@
+package main
```
DavisFrench, line 1: This is more of a nit pick than an actual issue, but I think auto_completion_test.go may be a better name for the file. I think it would be more in line with the naming convention already present in the repo.

DavisFrench (commented):
This looks really great, and I'm excited for this functionality to be added. I think that the only comment of mine that might need real consideration is the two tests that seemed to be similar in `cmd/ketch/builder_list_test.go`.

stinkyfingers (commented):
This looks and works great. I agree w/ Davis about the one test. After that, let's ship it.

cmd/ketch/autocompletion.go
koncar: ok, renamed :)

cmd/ketch/autocompletion_test.go
koncar: ok, renamed :)

cmd/ketch/builder_list_test.go
koncar: good catch, modified 2 tests

internal/deploy/parameters.go
koncar: i tried to follow pattern for now, we can plan future refactor of this as it's out of scope a bit

stinkyfingers (approved):
Nice work! Looks good and I tried it again successfully.

DavisFrench (approved):
All tests pass and the auto completion worked great for me on zsh.

koncar:
thx for reviews guys

PR #100, opened 2021-06-16
Hackathon deploy unit functionality

DavisFrench:
# Description

Merge the unit command functionality into an app deploy flag. The unit add and unit remove command functionality was already covered by the unit set command, as such I have opted only to include the unit set functionality in the units flag.

The ability to set units per version and or process has also been included with the optional unit-version and unit-process flags. Trying to use these flags without the units flag will result in an error.

Refactoring of the deployFromImage and deployFromSource functions into one function with a conditional used to determine whether or not to build the image from source.

Fixes #1525

## Type of change

## Testing

## Documentation

## Final Checklist:


internal/deploy/parameters.go
```diff
@@ -312,6 +331,43 @@ func (c *ChangeSet) getBuilder(spec ketchv1.AppSpec) string {
 	return *c.builder
 }
 
+func (c *ChangeSet) getUnits() (int, error) {
+	if c.units == nil {
+		return 0, nil
+	}
+	if *c.units < 1 {
```
stinkyfingers, line 338: What do we think about permitting setting units to 0, essentially stopping the App? Crazy?

stinkyfingers (approved):
Looks good. I asked maybe-terrible design question.

internal/deploy/parameters.go
DavisFrench: Kubernetes doesn't support stopping/pausing as far as I know, but I think setting replicas (pods) to 0 is a way to kind of get the same effect. As far as this pr is concerned, I think that it is a bit out of scope since it will require some more discussion.

internal/deploy/deploy.go
```diff
@@ -58,11 +58,7 @@ func (r Runner) Run(ctx context.Context, svc *Services) error {
 		return err
 	}
 
-	if r.params.sourcePath != nil {
-		return deployFromSource(ctx, svc, app, r.params)
-	}
-
-	return deployFromImage(ctx, svc, app, r.params)
+	return deployImage(ctx, svc, app, r.params)
```
stinkyfingers, line 61: Nice.

stinkyfingers (approved):
Looks good. I ran it a few times and things seem to work as expected.

aleksej-paschenko (commented):
Before we drop the `unit` commands, I think we need to be sure all use cases are covered. 

If I have an application with two processes and I deploy it from the app's source code with the following command:

```bash
ketch app deploy myapp --framework dev --image image:v1 <source-path>
```

How can I scale the app units without building a new image?

A command `ketch app deploy myapp --image image:v1 --units=3` doesn't work for me because ketch gets processes from a Procfile and it looks like I have to provide a directory with the source code. But then ketch will build a new image.

DavisFrench:
When providing a source directory a new image will always be created. To update the application that uses that image, simply remove the path from the deploy command, and provide any additional flags that you wish to use for updating. Also, when deploying from an image built by pack, the code will automatically detect the processes and run them w/o the need of having the original source code and rebuilding it.

cmd/ketch/app_deploy_test.go
```diff
@@ -150,6 +151,7 @@ kubernetes:
 web: python app.py
 worker: python app.py
 `
+	packBuildMetadata string = "{\"bom\":null,\"buildpacks\":[{\"id\":\"heroku/python\",\"version\":\"0.3.1\"},{\"id\":\"heroku/procfile\",\"version\":\"0.6.2\"}],\"launcher\":{\"version\":\"0.11.3\",\"source\":{\"git\":{\"repository\":\"github.com/buildpacks/lifecycle\",\"commit\":\"aa4bbac\"}}},\"processes\":[{\"type\":\"web\",\"command\":\"python app.py\",\"args\":null,\"direct\":false,\"buildpackID\":\"heroku/procfile\"},{\"type\":\"worker\",\"command\":\"python app.py\",\"args\":null,\"direct\":false,\"buildpackID\":\"heroku/procfile\"},{\"type\":\"worker1\",\"command\":\"python app.py\",\"args\":null,\"direct\":false,\"buildpackID\":\"heroku/procfile\"}]}"
 )
 
 func TestNewCommand(t *testing.T) {
```
aleksej-paschenko, line 157: I think there is an upcoming issue with these unit tests: they are not unit tests. There are tests that cover few cases for the top-level function and no tests that have good coverage for small functions.

aleksej-paschenko (changes requested):

cmd/ketch/app_deploy_test.go
```diff
@@ -564,6 +566,41 @@ func TestNewCommand(t *testing.T) {
...
+				"--framework", "myframework",
+				"--image", "shipa/go-sample:latest",
+				"--procfile", "./Procfile",
+			},
+			setup: func(t *testing.T) {
+				dir := t.TempDir()
+				require.Nil(t, os.Mkdir(path.Join(dir, "src"), 0700))
+				require.Nil(t, os.Chdir(dir))
+			},
+			validate: func(t *testing.T, m deploy.Client) {
```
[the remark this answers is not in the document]
DavisFrench: A type cast is required, but every single test that supplies a validate function has to do the type cast and confirm it works. This is a bit repetitive, so I moved the conversion to the if statement that checks for the presence of the validate function. This allowed me to change the function to signature to what you suggested.

cmd/ketch/app_deploy_test.go
DavisFrench: Do you mean testing the functions in deploy.go individually?

DavisFrench:
Completed in another pr

PR #102, opened 2021-06-16
Shipa 1461 set default builder

DavisFrench:
# Description

Allow for the user to set their own default builder to be used whenever a new application is created from source. The name of the builder is persisted via a config.toml. The default location of the config.toml is expected to be `~/.ketch/`, but this can be overwritten by exporting the path to config.toml as `KETCH_HOME`

Fixes #1461

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/cmd/ketch/builder_set.go
+++ b/cmd/ketch/builder_set.go
@@ -23,8 +23,8 @@ func newBuilderSetCmd(ketchConfig configuration.KetchConfig) *cobra.Command {
 	return cmd
 }
 
-func setDefaultBuilder(ketchConfig configuration.KetchConfig, defaultBulder string) error {
-	ketchConfig.DefaultBuilder = defaultBulder
+func setDefaultBuilder(ketchConfig configuration.KetchConfig, defaultBuilder string) error {
+	ketchConfig.DefaultBuilder = defaultBuilder
 	path, err := configuration.DefaultConfigPath()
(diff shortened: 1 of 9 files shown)


cmd/ketch/builder_set.go
```diff
@@ -0,0 +1,33 @@
...
+	return cmd
+}
+
+func setDefaultBuilder(ketchConfig configuration.KetchConfig, defaultBuilder string) error {
+	ketchConfig.DefaultBuilder = defaultBuilder
+	path, err := configuration.DefaultConfigPath()
+	if err != nil {
+		return err
+	}
+	return configuration.Write(ketchConfig, path)
```
stinkyfingers, line 32: This will overwrite any other data currently in the ketch-config file, e.g. additional-builders. Should we do a get & update thing here?

DavisFrench: It does overwrite in the sense that the whole config.toml is rewriiten, but the function takes in the current configuration and modifies just the default-builder before printing the config back out to the file.  This makes it so that there should be no loss of the user defined additional-builders when setting the default-builder.

stinkyfingers: I completely missed line 27. Ship it!

stinkyfingers (approved):

PR #97, opened 2021-06-15
Update dependencies

stinkyfingers:
# Description

Updates dependencies, notably `k8s.io` and `sigs.k8s.io`. In `sigs.k8s.io/controller-runtime`, the `Client` interface methods now take `client.Object` as a parameter instead of `runtime.Object` (the client Object inherits the runtime object). The `Reconciler` signature includes context as a parameter. Changes reflect these updates. 

Fixes # (issue)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Chore (documentation addition or typo, file relocation)

## Testing

- [ ] New tests were added with this PR that prove my fix is effective or that my feature works (describe below this bullet)
- [x] This change requires no testing (i.e. documentation update)

## Documentation

- [ ] All added public packages, funcs, and types have been documented with doc comments
- [ ] I have commented my code, particularly in hard-to-understand areas

## Final Checklist:

- [x] I followed standard [GitHub flow](https://guides.github.com/introduction/flow/) guidelines
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings

## Additional Information (omit if empty)

Please include anything else relevant to this PR that may be useful to know.

PR #105, opened 2021-06-23
Revert "Update dependencies"

stinkyfingers:
Reverts shipa-corp/ketch#97

Vivek & Davis found that updated dependencies broke `app deploy`. The app seems to stay in a Creating state. This quick revert fixes.


DavisFrench (approved):
I have verified that apps are being deployed both from source and image. Units and processes are also behaving as expected. Two processes with one unit a piece results in two pods, one for each process. Similarly a one process deployment with 3 units results in 3 pods.

PR #109, opened 2021-06-30
Fix procfile

DavisFrench:
# Description

Remove the procfile command line flag. There is a pack convention that mandates Procfile is in the source directory. I enforce this requirement on the ketch side of things too, and if Procfile is not found in the src path the user will receive an error.

When creating the chart.Procfile type, deploying from image depends on v1.ConfigFile for grabbing the entrypoints and cmds. However, when creating from source, the required Procfile is read in and parsed due to images built with pack config have a strange v1.ConfigFile get generated.

I have tested deploying the bulletin board image and the python samples and have confirmed they work.

Fixes #1569

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Testing

## Documentation

## Final Checklist:


internal/chart/procfile.go
```diff
@@ -52,8 +52,7 @@ func ParseProcfile(content string) (*Procfile, error) {
 	for _, process := range procfile {
 		if p := procfileRegex.FindStringSubmatch(process); p != nil {
 			name := p[1]
-			cmd := p[2]
-			processes[name] = []string{strings.TrimSpace(cmd)}
+			processes[name] = []string{strings.TrimSpace(name)}
```
stinkyfingers, line 55: This change seems to work, but even with the PR description, I'm still confused as to why we can disregard the Processes' Commands here.

DavisFrench: I'll add an additional comment to clear things up, since it isn't particularly intuitive. The Procfile that pack wants looks exactly like you would expect, something like
```
web: python app.py
worker: python app.py
```
The thing is, in the image that pack creates things look a bit different. The names specified in the Procfile end up becoming executables located under `/cnb/process`, so in this case we have `web` and `worker` within that folder. If I want to execute the commands specified in the Procfile I have to run them with either `web` or `worker` since they are in the path whereas something like python won't be.

aleksej-paschenko: if it's deployment from source, a process' command line has to match `name`. 
For deployment from an image, we provide the right command line
```
Application: myapp
Framework: dev
Address: http://myapp.10.98.0.185.shipa.cloud

No environment variables.
DEPLOYMENT VERSION    IMAGE                                        PROCESS NAME    WEIGHT    STATE        CMD
1                     docker.io/shipasoftware/bulletinboard:1.0    web             100%      1 running    docker-entrypoint.sh npm start

```
so it looks OK to me

aleksej-paschenko (approved):

stinkyfingers (approved):
Thanks for dropping that comment in there. That will keep people like me from accidentally undoing this work in the future.

PR #111, opened 2021-07-05
Fix weight calculation of a step for canary deployment

aleksej-paschenko:


itsksaurabh (approved):
LGTM. Just fix the failing tests.

PR #112, opened 2021-07-06
Framework add/update by yaml

stinkyfingers:
# Description
https://shipaio.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=SHIPA&modal=detail&selectedIssue=SHIPA-1546

Adds ability to add & update frameworks by yamll file, e.g. `ketch add framework myframework.yaml` and `ketch add framework ./mydir/myframework.yaml`.

Notes:
- If the user specifies a .yaml file as the arg to `framework add` or `framework update`, additional flags are ignored.
- During `framework add` ingress `ClassName` and `IngressType` are defaulted to `traefik` if the yaml files's fields are empty, similar to the flag-based approach. Fields `Version`, `AppQuotaLimit`, and `NamespaceName` also are defaulted if a value is not provided.
- I switched the `FrameworkSpec.AppQuotaLimit` field to a `*int`, so that we could determine during yaml unmarshaling if the field was zero or nil. The default is -1, so unmarshaling to a `0` when the field is unpopulated won't work. This change required dereferencing appQuotaLimit in a lot of places.

Fixes # 1546

## Type of change

## Testing

## Documentation

## Final Checklist:


internal/validation/validation.go
```diff
@@ -10,7 +10,8 @@ import (
 )
 
 var (
-	nameRegexp = regexp.MustCompile(`^[a-z][a-z0-9-]{0,39}$`)
+	nameRegexp         = regexp.MustCompile(`^[a-z][a-z0-9-]{0,39}$`)
+	yamlFilenameRegexp = regexp.MustCompile(`^[A-Za-z0-9-_\/\.\,]{1,255}\.(yaml|yml)$`)
```
aleksej-paschenko, line 14: this regexp doesn't work with non-English letters(

aleksej-paschenko (commented):
looks mostly done
several small things there
* Do we need to change the help strings of `ketch framework add/update`? 
No way to understand that both commands can consume yaml files. 

```bash
aleksej@zen:~/shipa-project/src/github.com/shipa-corp/ketch$ ./bin/ketch framework update -h

Update a framework.

Usage:
  ketch framework update FRAMEWORK [flags]

Flags:
      --app-quota-limit int               Quota limit for app when adding it to this framework
      --cluster-issuer string             ClusterIssuer to obtain SSL certificates
  -h, --help                              help for update
      --ingress-class-name string         if set, it is used as kubernetes.io/ingress.class annotations
      --ingress-service-endpoint string   an IP address or dns name of the ingress controller's Service
      --ingress-type ingress-type         ingress controller type: traefik or istio (default traefik)
      --namespace string                  Kubernetes namespace for this framework
```
* should we use the same approach as we use in shipa? 
  I mean `ketch framework config export -o filename`

internal/validation/validation.go
stinkyfingers: Discussed in Slack. Made the regex much more tolerant.

aleksej-paschenko (approved):

PR #113, opened 2021-07-07
Shipa 1525

DavisFrench:
Merge the unit command functionality into an app deploy flag. The unit add and unit remove command functionality was already covered by the unit set command, as such I have opted only to include the unit set functionality in the units flag.

The ability to set units per version and or process has also been included with the optional unit-version and unit-process flags. Trying to use these flags without the units flag will result in an error.

Refactoring of the deployFromImage and deployFromSource functions into one function with a conditional used to determine whether or not to build the image from source.

Fixes #1525

## Testing

## Documentation

## Final Checklist:

When providing a source directory a new image will always be created. To update the application that uses that image, simply remove the path from the deploy command, and provide any additional flags that you wish to use for updating. Also, when deploying from an image built by pack, the code will automatically detect the processes and run them w/o the need of having the original source code and rebuilding it.

In regards to the 3 test cases that were removed from the procfile_test, all of them are covered by pack. Pack will fail to build an image in the event that the Procfile is not provided or is in an illicit format. Likewise, pack will take care of ignoring any comments and only provides us with what we care about. Any commands that the user specifies w/in the procfile is there responsibility. The previous regex ^([A-Za-z0-9_-]+):\s*(.+)$ allowed for the user to specify anything after the name, so the user being responsible for properly defining their commands has not changed. The name is still being held to the same standard defined in the first capture group of the regex above.


internal/deploy/deploy.go
```diff
@@ -339,14 +300,52 @@ func updateAppCRD(ctx context.Context, svc *Services, appName string, args updat
 			return errors.Wrap(err, "could not get app to deploy %q", appName)
 		}
 
+		if len(updated.Spec.Deployments) > 1 && !updated.Spec.Canary.Active {
+			return errors.New("cannot have more than one deployment per app, unless canary")
+		}
+
+		// allow user to update units on canary deployments
```
aleksej-paschenko, line 307: :+1:

internal/deploy/deploy.go
```diff
@@ -339,14 +300,52 @@ func updateAppCRD(ctx context.Context, svc *Services, appName string, args updat
...
+				}
+			}
+
+			return svc.Client.Update(ctx, &updated)
+		}
+
+		// if the previous deployment's image is the same as the user provided image we want to reuse
+		// certain details like the number of units per process
+		var usePreviousDeploymentSpecs bool
+		if len(updated.Spec.Deployments) == 1 {
```
aleksej-paschenko, line 322: :+1:

aleksej-paschenko (approved):

PR #132, opened 2021-09-06
Use go1.17

aleksej-paschenko:


koncar (changes requested):
i believe there are some dependencies that can be cleaned

can you do
go mod tidy -go=1.17

koncar (approved):

PR #138, opened 2021-09-09
Define imagePullSecrets per application deployment

aleksej-paschenko:

## Testing

## Documentation

### DIFF
--- a/internal/api/v1beta1/app_types.go
+++ b/internal/api/v1beta1/app_types.go
@@ -97,6 +97,10 @@ type AppDeploymentSpec struct {
 	Labels          []Label           `json:"labels,omitempty"`
 	RoutingSettings RoutingSettings   `json:"routingSettings,omitempty"`
 	ExposedPorts    []ExposedPort     `json:"exposedPorts,omitempty"`
+
+	// ImagePullSecrets contains a list of secrets to pull the image of this deployment.
+	// If this list is defined, app.Spec.DockerRegistrySpec is not used.
+	ImagePullSecrets []v1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
 }
 
 // IngressSpec configures entrypoints to access an application.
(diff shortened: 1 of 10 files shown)


internal/api/v1beta1/app_types.go
```diff
@@ -97,6 +97,10 @@ type AppDeploymentSpec struct {
 	Labels          []Label           `json:"labels,omitempty"`
 	RoutingSettings RoutingSettings   `json:"routingSettings,omitempty"`
 	ExposedPorts    []ExposedPort     `json:"exposedPorts,omitempty"`
+
+	// ImagePullSecrets contains a list of secrets to pull the image of this deployment.
+	// If this list is defined, app.Spec.DockerRegistrySpec is not used.
+	ImagePullSecrets []v1.LocalObjectReference `json:"imagePullSecrets,omitempty"`
```
itsksaurabh, line 103: why do we need a list here? I think 1 deployment can have 1 image so 1 secret per deployment?

itsksaurabh: How are we going to map secret to specific deployment if there is a list of these secrets?

aleksej-paschenko: i'm following the same approach 
https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers

there are multiple containers, in ketch they point to one image but it might be changed in the future

itsksaurabh: Do we support multi-container deployments? 
In the case of multi-container deployment how it maps image pull secrets to specific images? Di you tested it with pod having more than one private image?

aleksej-paschenko: We don't support multi-container deployments yet but the idea is to use exactly the same format defined in `PodSpec`
I haven't tested it, it is a standard kubernetes feature. 

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet_pods.go#L889
https://github.com/openshift/origin/blob/master/vendor/k8s.io/kubernetes/pkg/kubelet/images/image_manager.go#L89

stinkyfingers (approved):

PR #142, opened 2021-09-10
Shipa 1885 - shipa.io Group and shipa Namespace

stinkyfingers:
# Description

Permits running a Ketch controller configured to use the `shipa.io` Group in the `shipa` namespace along with a "standard" Ketch controller configured to use the `theketch.io` Group in the `ketch-system` namespace. 
- Updates controller code to adjust Group & Namespace based on flag `group` (duplicated from #142) and flag `namespace`
- Updates controller markers to allow `shipa.io` group in rbac

Fixes # [1885](https://shipaio.atlassian.net/browse/SHIPA-1885)

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -40,22 +40,24 @@ var (
 	setupLog = ctrl.Log.WithName("setup")
 )
 
-func init() {
-	_ = clientgoscheme.AddToScheme(scheme)
-	_ = ketchv1.AddToScheme(scheme)
-	// +kubebuilder:scaffold:scheme
-}
-
 func main() {
 	var metricsAddr string
 	var enableLeaderElection bool
 	var disableWebhooks bool
+	var group string
+	var namespace string
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.BoolVar(&disableWebhooks, "disable-webhooks", false, "Disable webhooks.")
+	flag.StringVar(&group, "group", ketchv1.TheKetchGroup, "specify a non-default group")
+	flag.StringVar(&namespace, "namespace", controllers.KetchNamespace, "specify a non-default namespace")
 	flag.Parse()
+
+	_ = clientgoscheme.AddToScheme(scheme)
+	_ = ketchv1.AddToScheme(ketchv1.WithGroup(group))(scheme)
+	// +kubebuilder:scaffold:scheme
 
 	ctrl.SetLogger(zap.New(zap.UseDevMode(true)))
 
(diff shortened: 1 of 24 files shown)


aleksej-paschenko (changes requested):
I'm not sure about this approach. We'll end up with three instances of `config`:
1. ketch/ketch config
2. ketch/shipa config
3. shipa/shipa config, this is a copy of 2

Does it make sense to go with 1 and 3 only? Do we need to have a very specific configuration for shipa in this repository? https://github.com/shipa-corp/shipa/pull/927/files

Moreover, shipa config and ketch config are different. Several things:
1. Ketch works in its own namespace, it is `ketch-system` and we don't create a dedicated serviceaccount, both `leader-election-rolebinding` and `manager-rolebinding` use `default` serviceaccount in `ketch-system`. That's OK for ketch, but in shipa we have to create a new service account to run ketch. I think we are going to run ketch inside `shipa-system` namespace along with all other stuff we have there. 
2. ClusterRole and ClusterRoleBindings are cluster-scoped resources. We need to use unique names in shipa config.
3. At least in the beginning we should start with disabled webhooks. Shipa handles all cases validated by ketch's webhooks and managing our own instance of cert-manager can take some time. 

And we need to be able to change [this](https://github.com/shipa-corp/ketch/blob/main/internal/controllers/constants.go#L6) constant.

stinkyfingers:
@aleksej-paschenko I removed all the shipa config files (and created a PR here: https://github.com/shipa-corp/ketch-types/pull/2. Is that what we're looking for?) This PR now adds a `group` flag (duplicated from #142), a `namespace` flag, and permissions for `shipa.io`.

internal/controllers/app_controller.go
```diff
@@ -63,10 +63,10 @@ type Helm interface {
 	DeleteChart(appName string) error
 }
 
-// +kubebuilder:rbac:groups=theketch.io,resources=apps,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=theketch.io,resources=apps/status,verbs=get;update;patch
-// +kubebuilder:rbac:groups=theketch.io,resources=frameworks,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups=theketch.io,resources=frameworks/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=theketch.io;shipa.io,resources=apps,verbs=get;list;watch;create;update;patch;delete
```
aleksej-paschenko, line 66: I think it is OK to keep only theketch.io here because it used to generate config/rbac/* and we are good to convert the ketch's config to shipa's config manually or with a script.

cmd/manager/main.go
```diff
@@ -40,23 +40,25 @@ var (
...
 	var disableWebhooks bool
+	var group string
+	var namespace string
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.BoolVar(&disableWebhooks, "disable-webhooks", false, "Disable webhooks.")
+	flag.StringVar(&group, "group", ketchv1.TheKetchGroup, "specify a non-default group")
+	flag.StringVar(&namespace, "namespace", controllers.KetchNamespace, "specify a non-default namespace")
```
aleksej-paschenko, line 55: :+1:

config/rbac/role.yaml
```diff
@@ -219,6 +219,7 @@ rules:
   - patch
   - update
 - apiGroups:
+  - shipa.io
```
aleksej-paschenko, line 222: with this ketch will be able to work with shipa's apps, not sure this is something we need

aleksej-paschenko (approved):

PR #143, opened 2021-09-14
Uses Group as prefix for apps' labels

stinkyfingers:
# Description

Makes app's labels configurable with `--group`.

My editor's linter cleaned up a few stray spaces.

Fixes # [1889](https://shipaio.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=SHIPA&modal=detail&selectedIssue=SHIPA-1889)

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/internal/api/v1beta1/groupversion_info.go
+++ b/internal/api/v1beta1/groupversion_info.go
@@ -20,17 +20,41 @@ limitations under the License.
 package v1beta1
 
 import (
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/scheme"
 )
 
-var (
-	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "theketch.io", Version: "v1beta1"}
-
-	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
-	SchemeBuilder = &scheme.Builder{GroupVersion: GroupVersion}
-
-	// AddToScheme adds the types in this group-version to the given scheme.
-	AddToScheme = SchemeBuilder.AddToScheme
+const (
+	TheKetchGroup = "theketch.io"
 )
+
+// Group is the exported variable indicating group, defaults to theketch.io
+var Group = TheKetchGroup
+
+type SchemeOptions struct{ group string }
+
+type schemeOption func(opts *SchemeOptions)
+
+func WithGroup(group string) schemeOption {
+	return func(opts *SchemeOptions) { opts.group = group }
+}
+func defaultSchemeOptions() SchemeOptions {
+	return SchemeOptions{group: TheKetchGroup}
+}
+
+// AddToScheme can be easily consumed by ketch-cli and by default it uses `theketch.io` // and it can be consumed by shipa:
+// AddToScheme(WithGroup("shipa.io"))
+func AddToScheme(opts ...schemeOption) func(s *runtime.Scheme) error {
+	options := defaultSchemeOptions()
+	for _, o := range opts {
+		o(&options)
+	}
+	groupVersion := schema.GroupVersion{Group: options.group, Version: "v1beta1"}
+	builder := &scheme.Builder{GroupVersion: groupVersion}
+	builder.Register(&App{}, &AppList{})
+	builder.Register(&Framework{}, &FrameworkList{})
+	builder.Register(&Job{}, &JobList{})
+	Group = options.group
+	return builder.AddToScheme
+}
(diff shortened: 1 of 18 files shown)


internal/api/v1beta1/groupversion_info.go
```diff
@@ -33,4 +35,15 @@ var (
...
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
+
+	// Group is the CRD group
+	Group = "theketch.io"
 )
+
+func init() {
+	if group := os.Getenv("CUSTOM_GROUP"); group != "" {
+		Group = group
```
DavisFrench, line 45: Just want to check, but should we confirm that the group pulled from the environment conforms to k8s standards?

DavisFrench (approved):
I think everything looks good, but there is one thing I want to confirm. Does `/internal/templates/job/yamls/job.yaml` also need `{{ $.Values.app.group }}` before app-name in the metadata labels?

stinkyfingers:
@DavisFrench 

> Does /internal/templates/job/yamls/job.yaml also need {{ $.Values.app.group }} before app-name in the metadata labels?

I don't _think_ so. I had it, then removed it as Shipa's ketch won't use `Jobs` for anything.

internal/api/v1beta1/groupversion_info.go
```diff
@@ -33,4 +35,15 @@ var (
 
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
+
+	// Group is the CRD group
+	Group = "theketch.io"
 )
+
+func init() {
```
aleksej-paschenko, line 43: I'm a bit concerned with such an initialization:
1. both ketch-cli and ketch-manager import this package, so the CUSTOM_GROUP variable works for both. Let's avoid this behavior. Ketch-cli should always work with the `theketch.io` group.
2. Shipa imports this package too and the only way to set `shipa.io` group is to set the CUSTOM_GROUP env, this looks tricky to me
3. A lot of initialization is done in [main.go](https://github.com/shipa-corp/ketch/blob/main/cmd/manager/main.go#L49), and usually it is a good point to understand what the parameters to configure ketch are. For example, one can open main.go and see that webhooks can be disabled with `--disable-webhooks`. The suggestion is to use CUSTOM_GROUP in main.go

so can we go with something like this:
// groupversion_info.go
```golang
const (
	TheKetchGroup = "theketch.io"
)

type SchemeOptions struct {
	group string
}

type SchemeOption func(opts *SchemeOptions)

func WithGroup(group string) SchemeOption {
	return func(opts *SchemeOptions) {
		opts.group = group
	}
}

func defaultSchemeOptions() SchemeOptions {
	return SchemeOptions{group: TheKetchGroup}
}

// AddToScheme can be easily consumed by ketch-cli and by default it uses `theketch.io`
// and it can be consumed by shipa: AddToScheme(WithGroup("shipa.io")) 

func AddToScheme(opts ...SchemeOption) func(s *runtime.Scheme) error {
	options := defaultSchemeOptions()
	for _, o := range opts {
		o(&options)
	}
	groupVersion := schema.GroupVersion{Group: options.group, Version: "v1beta1"}
	builder := &scheme.Builder{GroupVersion: groupVersion}
	return builder.AddToScheme
}

```

// main.go
```golang
func main() {
	_ = clientgoscheme.AddToScheme(scheme)

	group := ketchv1.TheKetchGroup
	if len(os.Getenv("CUSTOM_GROUP")) > 0 {
		group = os.Getenv("CUSTOM_GROUP")
	}
	_ = ketchv1.AddToScheme(ketchv1.WithGroup(group))(scheme)

		// +kubebuilder:scaffold:scheme
	var metricsAddr string
	var enableLeaderElection bool
	var disableWebhooks bool
	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
	flag.BoolVar(&enableLeaderElection, "enable-leader-election", true,
		"Enable leader election for controller manager. "+
			"Enabling this will ensure there is only one active controller manager.")
	flag.BoolVar(&disableWebhooks, "disable-webhooks", false, "Disable webhooks.")
	flag.Parse()
```

I think we should avoid complex `init` functions(this one takes arguments from env variable and has side-effects) and global variables (`Group`) in general

aleksej-paschenko (changes requested):

internal/api/v1beta1/groupversion_info.go
stinkyfingers: @aleksej-paschenko Thanks for the help. I moved the group logic out of `init()` and into `main()` as you suggested and switched from an environment variable to a `flag`. I'm curious as to how we'll end up implementing this - does a flag work?

aleksej-paschenko: it should work, we can add the flags there https://github.com/shipa-corp/shipa/pull/927/files#diff-f305fab607f0d165929a8f357219b61fe04fcae48fc9cdcb09ef1e86fad3c3c0R25

aleksej-paschenko (approved):

PR #150, opened 2021-09-28
[SHIPA-1944] Ketch to provides way to use user's SSL certificate for cnames

stinkyfingers:
# Description

Fixes # [1944](https://shipaio.atlassian.net/browse/SHIPA-1944) Ketch to provide way to use user's SSL certificate for cnames

- Assumes that an SSL Cert will be stored as a Secret. 
- Adds a `SecretNames` field per `App`.
- If `App` contains `SecretNames`, it creates ad hoc `clusterIssuer`s and overrides the default `framework.IngressController.ClusterIssuer` with the one created. This `clusterIssuer` is passed to traefik/istio. 

**See below for a lengthy description on testing this**

## Type of change

## Testing

## Documentation

## Final Checklist:

## Testing 
1. Create cert signing request config file named `san.cnf` (note: v3_ca required on OSX):
```
[ req ]

default_bits        = 2048
default_keyfile     = server-key.pem
distinguished_name  = subject
req_extensions      = req_ext
x509_extensions     = x509_ext
string_mask         = utf8only


[ subject ]

countryName                 = US
countryName_default         = US

stateOrProvinceName         = WI
stateOrProvinceName_default = NY

localityName                = EC
localityName_default        = New York

organizationName            = Shipa
organizationName_default    = Example, LLC

commonName                  = shipa.io
commonName_default          = Example Company

emailAddress                = jshenk@shipa.io
emailAddress_default        = test@example.com

[ x509_ext ]

subjectKeyIdentifier   = hash
authorityKeyIdentifier = keyid,issuer

basicConstraints       = CA:FALSE
keyUsage               = digitalSignature, keyEncipherment
subjectAltName         = @alternate_names
nsComment              = "OpenSSL Generated Certificate"

[ req_ext ]

subjectKeyIdentifier = hash

basicConstraints     = CA:FALSE
keyUsage             = digitalSignature, keyEncipherment
subjectAltName       = @alternate_names
nsComment            = "OpenSSL Generated Certificate"

[ alternate_names ]

DNS.1       = foo.shipa.cloud

[ v3_ca ]
basicConstraints = critical,CA:TRUE
subjectKeyIdentifier = hash
authorityKeyIdentifier = keyid:always,issuer:always
```

2. Create key/cert from csr (via config, above):
`openssl req -config san.cnf -new -sha256 -newkey rsa:2048 -nodes -keyout key.pem -x509 -days 365 -out cert.pem -extensions v3_ca`
This creates `key.pem` and self-signed `cert.pem` 

3. Create `secret.yaml` (or two) in k8s cluster:
```
apiVersion: v1
kind: Secret
metadata:
  name: foo.shipa.cloud
  namespace: cert-manager
type: kubernetes.io/tls
data:
  tls.key: >-
    <key.pem, base64 encoded>

  tls.crt: >-
    <cert.pem, base64 encoded>
```

`k apply -f secret.yaml`
NOTE: secret should be in same namespace as cert-manager pod

4. Create framework without clusterissuer, for example: 
```
apiVersion: theketch.io/v1beta1
kind: Framework
metadata:
  name: test
spec:
  namespace: ketch-test
  version: v1
  name: test
  appQuotaLimit: -1
  ingressController:
    type: traefik
    className: traefik
    serviceEndpoint: 10.96.13.219 (traefik's Loadbalancer IP)
```
`k apply -f framework.yaml` 

5. Create app with field `secretNames: foo.shipa.io` and a secure cname, e.g.
```
apiVersion: theketch.io/v1beta1
kind: App
metadata:
  name: test
  namespace: foobar
spec:
  canary: {}
  deployments:
  - image: gcr.io/kubernetes-312803/sample-go-app:latest
    processes:
    - cmd:
      - web
      name: web
      units: 1
    routingSettings:
      weight: 100
    version: 1
    labels:
      - name: test-label
        value: works
  deploymentsCount: 1
  dockerRegistry: {}
  framework: test
  ingress:
    generateDefaultCname: true
    cnames:
      - name: secure.10.96.13.219.shipa.cloud (NOTE: your Loadbalancer IP)
        secure: true
  secretNames: 
    - foo.shipa.cloud
    - ANY OTHER SECRETS YOU WANT TO TRY
```
`k apply -f app.yaml`

6.  Check things
- `curl https://secure.10.96.13.219.shipa.cloud --cacert cert.pem` (note: your loadbalancer IP)
- secure endpoint should not work w/o --cacert
- should not work if secret is bad/nonexistent (try altering the name of the app.SecretNames to a nonexistent secret)
- `k describe clusterissuer foo.shipa.io` to inspect the clusterissuers (one per secret) - will error if secret is bad or misnamed.

### DIFF
--- a/internal/api/v1beta1/app_types.go
+++ b/internal/api/v1beta1/app_types.go
@@ -214,7 +214,7 @@ type AppSpec struct {
 	Annotations []MetadataItem `json:"annotations,omitempty"`
 
 	// SecretNames are the secret TLS certifificates to associate with the App
-	SecretNames []string `json:"secretName,omitempty"`
+	SecretNames []string `json:"secretNames,omitempty"`
 }
 
 // MetadataItem represent a request to add label/annotations to processes
--- a/internal/api/v1beta1/app_types.go
+++ b/internal/api/v1beta1/app_types.go
@@ -213,8 +213,8 @@ type AppSpec struct {
 	// Annotations is a list of annotations that will be applied to processes
 	Annotations []MetadataItem `json:"annotations,omitempty"`
 
-	// Secret is the secret TLS certifificate to associate with the App
-	SecretName string `json:"secretName,omitempty"`
+	// SecretNames are the secret TLS certifificates to associate with the App
+	SecretNames []string `json:"secretName,omitempty"`
 }
 
 // MetadataItem represent a request to add label/annotations to processes
--- a/internal/api/v1beta1/app_types.go
+++ b/internal/api/v1beta1/app_types.go
@@ -52,7 +52,13 @@ type Label struct {
 }
 
 // CnameList is a list of an app's CNAMEs.
-type CnameList []string
+type CnameList []Cname
+
+// Cname represents a DNS record and whether the record use TLS.
+type Cname struct {
+	Name   string `json:"name"`
+	Secure bool   `json:"secure"`
+}
 
 // RoutingSettings contains a weight of the current deployment used to route incoming traffic.
 // If an application has two deployments with corresponding weights of 30 and 70,
@@ -206,6 +212,9 @@ type AppSpec struct {
 
 	// Annotations is a list of annotations that will be applied to processes
 	Annotations []MetadataItem `json:"annotations,omitempty"`
+
+	// Secret is the secret TLS certifificate to associate with the App
+	SecretName string `json:"secretName,omitempty"`
 }
 
 // MetadataItem represent a request to add label/annotations to processes
@@ -379,17 +388,17 @@ func (app *App) Start(selector Selector) error {
 
 // CNames returns all CNAMEs to access the application including a default cname.
 func (app *App) CNames(framework *Framework) []string {
-	scheme := "http"
-	if len(framework.Spec.IngressController.ClusterIssuer) > 0 {
-		scheme = "https"
-	}
 	cnames := []string{}
 	defaultCname := app.DefaultCname(framework)
 	if defaultCname != nil {
 		cnames = append(cnames, fmt.Sprintf("http://%s", *defaultCname))
 	}
 	for _, cname := range app.Spec.Ingress.Cnames {
-		cnames = append(cnames, fmt.Sprintf("%s://%s", scheme, cname))
+		scheme := "http"
+		if cname.Secure {
+			scheme = "https"
+		}
+		cnames = append(cnames, fmt.Sprintf("%s://%s", scheme, cname.Name))
 	}
 	return cnames
 }
--- a/internal/chart/application_chart.go
+++ b/internal/chart/application_chart.go
@@ -18,6 +18,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/yaml"
 
+	"github.com/pkg/errors"
 	ketchv1 "github.com/shipa-corp/ketch/internal/api/v1beta1"
 	"github.com/shipa-corp/ketch/internal/templates"
 )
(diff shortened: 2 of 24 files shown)


internal/chart/testdata/charts/dashboard-traefik-secret.yaml
```diff
@@ -0,0 +1,387 @@
+---
```
stinkyfingers, line 1: This test file is the same as `dashboard-traefik` with addition of a `ClusterIssuer` at line 339, which is generated because the tested `App` has a populated `SecretName`.

internal/chart/application_chart.go
```diff
@@ -70,8 +73,8 @@ type app struct {
 	// For example, "spec.rules" of an Ingress object must contain at least one rule.
 	IsAccessible bool   `json:"isAccessible"`
 	Group        string `json:"group"`
-
-	Service *gatewayService
+	Secret       string `json:"secret"`
```
itsksaurabh, line 76: not sure but are "secretName" and json:"secret" same things? 
If yes how about have similar naming for both of them? :)

internal/chart/application_chart.go
```diff
@@ -329,33 +333,39 @@ func isAppAccessible(a *app) bool {
...
 	var http []string
-	var https []string
+	var https []httpsEndpoint
 
 	for _, cname := range app.Spec.Ingress.Cnames {
 		if cname.Secure {
-			if len(framework.Spec.IngressController.ClusterIssuer) == 0 {
-				return nil, errors.New("secure cnames require a framework.Ingress.ClusterIssuer to be specified")
+			secretName := app.Spec.SecretName
+			clusterIssuer := fmt.Sprintf("%s-clusterissuer", app.Spec.SecretName)
```
itsksaurabh, line 341: do we need to initialize them inside this loop for every iteration? Can they be outside the loop?

internal/templates/common/yamls/clusterIssuer.yaml
```diff
@@ -0,0 +1,10 @@
+{{- if $.Values.app.secret }}
+apiVersion: cert-manager.io/v1
+kind: ClusterIssuer
+metadata:
+  name: {{ $.Values.app.secret }}-clusterissuer
+spec:
+  ca:
+    secretName: {{ $.Values.app.secret }}
```
itsksaurabh, line 8: When referencing a Secret resource in ClusterIssuer resources, the Secret needs to be in the same namespace as the cert-manager controller pod. Do we have some checks for it already?

We can optionally override this by using the --cluster-resource-namespace argument to the controller but not sure if we are going to do it.

itsksaurabh (changes requested):

internal/api/v1beta1/app_types.go
```diff
@@ -212,6 +212,9 @@ type AppSpec struct {
 
 	// Annotations is a list of annotations that will be applied to processes
 	Annotations []MetadataItem `json:"annotations,omitempty"`
+
+	// Secret is the secret TLS certifificate to associate with the App
+	SecretName string `json:"secretName,omitempty"`
```
aleksej-paschenko, line 217: is it possible to use an array of secrets?

aleksej-paschenko (changes requested):

internal/chart/application_chart.go
stinkyfingers: Good call. Changed to `secretName`.

internal/chart/application_chart.go
stinkyfingers: I cleaned this up as part of enabling multiple secrets.

internal/templates/common/yamls/clusterIssuer.yaml
stinkyfingers: Good call. I added a check for the secret (in ns `cert-manager`) to the reconciler.

internal/api/v1beta1/app_types.go
stinkyfingers: You bet. I made the change and tested out multiple secrets.

itsksaurabh (approved):
Looks awesome !

stinkyfingers:
See https://github.com/theketchio/ketch/pull/164

PR #155, opened 2021-09-30
Put values of annotations and labels in quotes

aleksej-paschenko:
Helm fails to validate the metadata section of a k8s resource
   if there is a label or annotations without quotes like in the following example

   apiVersion: apps/v1
   kind: Deployment
   metadata:
     annotations:
       name: true

### DIFF
--- a/internal/templates/common/yamls/service.yaml
+++ b/internal/templates/common/yamls/service.yaml
@@ -11,15 +11,15 @@ metadata:
     {{ $.Values.app.group }}/app-deployment-version: {{ $deployment.version | quote }}
     {{ $.Values.app.group }}/is-isolated-run: "false"
     {{- range $i, $label := $deployment.labels }}
-    {{ $label.name }}: {{ $label.value }}
+    {{ $label.name }}: "{{ $label.value }}"
     {{- end }}
     {{- range $k, $v := $process.extra.serviceMetadata.labels }}
-    {{ $k }}: {{ $v }}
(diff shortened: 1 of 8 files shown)


internal/templates/common/yamls/service.yaml
```diff
@@ -11,15 +11,15 @@ metadata:
     {{ $.Values.app.group }}/app-deployment-version: {{ $deployment.version | quote }}
     {{ $.Values.app.group }}/is-isolated-run: "false"
     {{- range $i, $label := $deployment.labels }}
-    {{ $label.name }}: {{ $label.value }}
+    {{ $label.name }}: "{{ $label.value }}"
     {{- end }}
     {{- range $k, $v := $process.extra.serviceMetadata.labels }}
-    {{ $k }}: {{ $v }}
+    {{ $k }}: "{{ $v }}"
```
itsksaurabh, line 17: would it be possible to use helm pipeline function here?

https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#pipelines

```
........ | quote }}
```

itsksaurabh: same inside everywhere we used hardcoded " "?

itsksaurabh (approved):

internal/templates/common/yamls/service.yaml
aleksej-paschenko: the resulting yamls are valid for kubernetes, helm validates the `metadata` section and that's the issue.

PR #164, opened 2021-10-13
[SHIPA-1944] Provide way to use user's SSL certificate for cnames

aleksej-paschenko:
This PR adds a new `secretName` field to Ketch App CRD. If provided, ketch doesn't obtain an SSL certificate for such a cname but expects that a given secret contains a valid SSL certificate. 

```yaml
spec:
  ingress:
    cnames:
    - name: app.shipa.io
      secure: true
    - name: dashboard.shipa.io
      secretName: my-secret
      secure: true
```

## Type of change

## Testing

### DIFF
--- a/internal/chart/application_chart_test.go
+++ b/internal/chart/application_chart_test.go
@@ -134,13 +134,9 @@ func TestNewApplicationChart(t *testing.T) {
 			Ingress: ketchv1.IngressSpec{
 				GenerateDefaultCname: true,
 				Cnames: []ketchv1.Cname{
-					{
-						Name:   "theketch.io",
-						Secure: true,
-					}, {
-						Name:   "app.theketch.io",
-						Secure: true,
-					},
+					{Name: "theketch.io", Secure: true},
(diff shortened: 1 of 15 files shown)


internal/chart/application_chart_test.go
```diff
@@ -298,7 +294,7 @@ func TestNewApplicationChart(t *testing.T) {
 			require.Nil(t, err)
 
 			expectedFilename := filepath.Join(chartDirectory, fmt.Sprintf("%s.yaml", tt.wantYamlsFilename))
-			actualFilename := filepath.Join(chartDirectory, fmt.Sprintf("%s-output.yaml", tt.wantYamlsFilename))
+			actualFilename := filepath.Join(chartDirectory, fmt.Sprintf("%s.output.yaml", tt.wantYamlsFilename))
```
aleksej-paschenko, line 297: it's easier to work with all these files because they are sorted in a way that <file.yaml> and <file.output.yaml> are neighbors.

PR #192, opened 2021-11-05
Post render

aleksej-paschenko:


DavisFrench:
If the framework where the app is being deployed to contains a configmap that has a name like `some-name-postrender` (the `postrender` being present is the important part), the contents of that config map will be applied to the apps helmchart before being deployed.

The config map itself should contain a `kustomization.yaml` key with its corresponding contents as the value, and any other needed yaml files (e.g. `patch.yaml`) as keys with the contents as the value. These yamls should be under the configmap's data key.

internal/chart/helm_client.go
```diff
@@ -51,9 +61,75 @@ func getActionConfig(namespace string) (*action.Configuration, error) {
...
 
+type postRender struct {
+	client.Client
+	namespace string
+}
+
+func (p *postRender) Run(renderedManifests *bytes.Buffer) (modifiedManifests *bytes.Buffer, err error) {
+
+	var configMapList v1.ConfigMapList
+	if err := p.Client.List(context.Background(), &configMapList); err != nil {
```
aleksej-paschenko, line 75: How about getting configmaps from the namespace we are interested in? 
```golang

	var configMapList v1.ConfigMapList
	opts := &client.ListOptions{Namespace: p.namespace}
	if err := p.Client.List(context.Background(), &configMapList, opts); err != nil {
		return nil, err
	}

```

internal/chart/helm_client.go
```diff
@@ -51,9 +61,75 @@ func getActionConfig(namespace string) (*action.Configuration, error) {
 	return actionConfig, nil
 }
 
-// Option to perform additional configuration of action.Install before running a chart installation.
+// InstallOption to perform additional configuration of action.Install before running a chart installation.
 type InstallOption func(install *action.Install)
 
+type postRender struct {
+	client.Client
```
aleksej-paschenko, line 68: why do we want to embed client.Client? 
how about 
```golang
type postRender struct {
	cli client.Client
	namespace string
}

```

internal/chart/helm_client.go
```diff
@@ -51,9 +61,75 @@ func getActionConfig(namespace string) (*action.Configuration, error) {
...
+
+	fs := filesys.MakeFsInMemory()
+	if err := fs.Mkdir(p.namespace); err != nil {
+		return nil, err
+	}
+
+	var postrenderFound bool
+	for _, cm := range configMapList.Items {
+		// only apply the postrenders present in the current namespace
+		if cm.Namespace == p.namespace {
```
aleksej-paschenko, line 87: I think it would be nice if we could try to decrease quantity of all of these nested statements.
```golang
	var postrenderFound bool
	for _, cm := range configMapList.Items {
		if !strings.HasSuffix(cm.Name, "-postrender") {
			continue
		}
		// check if a postrender is available
		postrenderFound = true
		for k, v := range cm.Data {
			fileName := p.namespace + "/" + k
			if err := fs.WriteFile(fileName, []byte(v)); err != nil {
				return nil, err
			}
		}
	}


```

internal/chart/helm_client.go
```diff
@@ -51,9 +61,75 @@ func getActionConfig(namespace string) (*action.Configuration, error) {
 	return actionConfig, nil
 }
 
-// Option to perform additional configuration of action.Install before running a chart installation.
+// InstallOption to perform additional configuration of action.Install before running a chart installation.
 type InstallOption func(install *action.Install)
 
+type postRender struct {
```
aleksej-paschenko, line 67: let's move it to `post_render.go`?

aleksej-paschenko (commented):
looks awesome!

internal/chart/helm_client.go
DavisFrench: Good idea, added.

internal/chart/helm_client.go
DavisFrench: Changed.

internal/chart/helm_client.go
DavisFrench: Changed.

internal/chart/helm_client.go
DavisFrench: Done.

config/manager/kustomization.yaml
```diff
@@ -4,4 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: gcr.io/kubernetes-312803/ketch
+  newName: davisdavis27french/post
```
stinkyfingers, line 7: 🧇

internal/chart/helm_client.go
```diff
@@ -51,9 +53,11 @@ func getActionConfig(namespace string) (*action.Configuration, error) {
 	return actionConfig, nil
 }
 
-// Option to perform additional configuration of action.Install before running a chart installation.
+// InstallOption to perform additional configuration of action.Install before running a chart installation.
 type InstallOption func(install *action.Install)
 
+var _ postrender.PostRenderer = &postRender{}
```
aleksej-paschenko, line 59: can we either remove it or move to `post_render.go`?

aleksej-paschenko (commented):
looks great, would it be much work to add a test?

internal/chart/helm_client.go
DavisFrench: Moved.

config/manager/kustomization.yaml
DavisFrench: Changed back.

config/manager/kustomization.yaml
```diff
@@ -5,3 +5,4 @@ kind: Kustomization
 images:
 - name: controller
   newName: gcr.io/kubernetes-312803/ketch
+  newTag: "20"
```
aleksej-paschenko, line 8: do we need this?

DavisFrench: No, I'll delete it real quick.

aleksej-paschenko:
Looks awesome!

aleksej-paschenko:
@stinkyfingers if it works for you, let's approve it and merge, I can't approve this PR because it was opened by me

internal/chart/post_render.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"helm.sh/helm/v3/pkg/postrender"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/kustomize/api/krusty"
+	kTypes "sigs.k8s.io/kustomize/api/types"
+	"sigs.k8s.io/kustomize/kyaml/filesys"
+)
+
+var _ postrender.PostRenderer = &postRender{}
```
stinkyfingers, line 16: Is this anonymous var needed? What's it do?

internal/chart/post_render_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+//go:embed testdata/render_yamls/prerendered-manifests.yaml
```
stinkyfingers, line 15: 🆒

stinkyfingers (approved):

internal/chart/post_render.go
stinkyfingers: Davis reminded my of implicit interface implementation via Slack.

DavisFrench: This is used to "enforce" that postRender implements the interface postrender.PostRenderer. It will give a more verbose error when compiling if the interface is not implemented. https://stackoverflow.com/questions/33089523/how-to-mark-golang-struct-as-implementing-interface

PR #211, opened 2021-12-21
Keep only 1 historical helm release per app on cluster

aleksej-paschenko:
Currently, we don't specify MaxHistory and a number of helm releases
keeps growing meaning ketch creates lots of secrets named
`sh.helm.release.v1.<app-name>.v<version>`.

## Type of change

### DIFF
--- a/internal/chart/helm_client.go
+++ b/internal/chart/helm_client.go
@@ -1,12 +1,15 @@
 package chart
 
 import (
+	"errors"
+	"log"
+	"os"
+
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/release"
+	"helm.sh/helm/v3/pkg/storage/driver"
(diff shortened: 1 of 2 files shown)


internal/chart/helm_client.go
```diff
@@ -91,6 +94,10 @@ func (c HelmClient) UpdateChart(tv TemplateValuer, config ChartConfig, opts ...I
 	}
 	updateClient := action.NewUpgrade(c.cfg)
 	updateClient.Namespace = c.namespace
+	// MaxHistory=0 means there is an unlimited number of k8s secrets per application,
```
itsksaurabh, line 97: We might need to update comments, we are only mentioning `secrets` here but helm docs say:

```
MaxHistory specifies the maximum number of historical releases that will be retained, including the most recent release.
Values of 0 or less are ignored (meaning no limits are imposed).

```

itsksaurabh (approved):
Looks great just added a minor suggestion. :)

stinkyfingers (approved):

PR #223, opened 2022-01-17
Dont create events when ketch controller gets a conflict error

aleksej-paschenko:
Ketch controller will eventually update an application CR and we are
good to print this type of errors to logs only.

### DIFF
--- a/internal/controllers/app_controller.go
+++ b/internal/controllers/app_controller.go
@@ -30,6 +30,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	"k8s.io/api/autoscaling/v2beta1"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


internal/controllers/app_controller.go
```diff
@@ -192,73 +201,75 @@ func hpaTargetMap(app *ketchv1.App, hpaList v2beta1.HorizontalPodAutoscalerList)
...
 }
 
-func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App, logger logr.Logger) reconcileResult {
+type appReconcileResult struct {
+	framework  *v1.ObjectReference
+	useTimeout bool
+	err        error
+}
+
+func (r appReconcileResult) isConflictError() bool {
```
koncar, line 210: could we add docs, to explain what it does, and why it has for {}

koncar (approved):

PR #250, opened 2022-05-12
[SHIPA-2788] Annotates events with Pod Name in the event of a Error type event

stinkyfingers:
# Description

Shipa would like to emit app.error Shipa events when there is a deployment error via kubernetes. It would be helpful to have one of the offending pods for this. This PR uses the existing `getPodStatus` func to get the podName, which is then included in the K8s event as an annotation. 

Fixes # shipa-2788

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/internal/controllers/app_controller.go
+++ b/internal/controllers/app_controller.go
@@ -271,6 +271,7 @@ func (cli workloadClient) Get(ctx context.Context) (*workload, error) {
 		if err != nil {
 			return nil, err
 		}
+		fmt.Println("...workload status", o.Status)
 		w := workload{
 			Name:               o.Name,
 			UpdatedReplicas:    int(o.Status.UpdatedReplicas),
@@ -372,7 +373,7 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App, logger 
 		}
 
 		// retry until all pods for canary deployment comes to running state.
-		if err := checkPodStatus(r.Group, r.Client, app.Name, app.Spec.Deployments[1].Version); err != nil {
+		if _, err := checkPodStatus(r.Group, r.Client, app.Name, app.Spec.Deployments[1].Version); err != nil {
 
 			if !timeoutExpired(app.Spec.Canary.Started, r.Now()) {
 				return appReconcileResult{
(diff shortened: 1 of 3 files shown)


internal/controllers/app_controller.go
```diff
@@ -271,6 +271,7 @@ func (cli workloadClient) Get(ctx context.Context) (*workload, error) {
 		if err != nil {
 			return nil, err
 		}
+		fmt.Println("...workload status", o.Status)
```
kavinaravind, line 274: did ya forget to remove this print statement?

internal/controllers/app_controller.go
```diff
@@ -372,7 +373,7 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App, logger
 		}
 
 		// retry until all pods for canary deployment comes to running state.
-		if err := checkPodStatus(r.Group, r.Client, app.Name, app.Spec.Deployments[1].Version); err != nil {
+		if _, err := checkPodStatus(r.Group, r.Client, app.Name, app.Spec.Deployments[1].Version); err != nil {
```
kavinaravind, line 376: I see the `len(app.Spec.Deployments) <= 1` check right before this call, is there ever a case where Deployments is 0 and `app.Spec.Deployments[1].version` would result in a null pointer or as its a canary deployment we are okay?

internal/controllers/app_controller.go
```diff
@@ -583,13 +584,15 @@ func (r *AppReconciler) watchFunc(ctx context.Context, cleanup cleanupFunc, app
 				recorder.AnnotatedEventf(app, appDeploymentEvent.Annotations, v1.EventTypeNormal, ketchv1.AppReconcileUpdate, appDeploymentEvent.Description)
 			}
 		case <-healthcheckTimeout:
+			podName, _ := checkPodStatus(r.Group, r.Client, app.Name, app.Spec.Deployments[len(app.Spec.Deployments)-1].Version)
```
kavinaravind, line 587: more for learning on my side, but why is there no need to check if error != nil for the following cases?

kavinaravind (changes requested):
small items and learning on my side

internal/controllers/app_controller.go
stinkyfingers: You don't want to read all my debugs?

internal/controllers/app_controller.go
stinkyfingers: I added a len check. I don't see where we assure that there are deployments & assume it can't hurt to check.

kavinaravind (approved):

PR #255, opened 2022-05-31
App post renderer

koncar:
# Description

Add the ability to have multiple post renders

App postrender is done per application deployment version

## Type of change

## Testing

## Documentation

## Final Checklist:


internal/chart/post_render.go
```diff
@@ -27,51 +33,68 @@ func (p *postRender) Run(renderedManifests *bytes.Buffer) (modifiedManifests *by
...
-		return nil, err
-	}
-
-	var postrenderFound bool
+	finalBuffer := renderedManifests
 	for _, cm := range configMapList.Items {
-		if !strings.HasSuffix(cm.Name, "-postrender") {
+		fwPatch := strings.HasSuffix(cm.Name, "-postrender")
+		var appPatch bool
+		for _, dv := range p.deploymentVersions {
```
aleksej-paschenko, line 40: A map looks like a better fit for `deploymentVersions`: 
1. no need to have this `range` loop
2. a map is a set, no order expectations and so on

aleksej-paschenko (approved):
looks great

internal/chart/post_render.go
koncar: in that case we would need to split the string and check if there is some deployment version inside there. it's same in terms of perfomance, and we are keeping map with empty struct which could be confusing

PR #256, opened 2022-06-02
Move application readiness wait loop inside goroutine

stinkyfingers:
# Description

Ketch gets stuck in a 10 minute loop when we deploy a `statefulset` with an error. Example `ketch app deploy testapp -k dev -i docker.io/shipasoftware/bulletinboard:1.0 --volume aaa/bbb` (malformed volume).

Unlike `deployments`, a `statefulset` with this error (and possibly others) never reaches a first `status.observedGeneration`. As a result, we sit in a loop waiting to proceed to kicking off the eventWatcher goroutine (watches Pod Events and emits Shipa Events). This blocks the reconciler, making it difficult for users to remove the bad application.

This PR moves the `observedGeneration` wait inside the goroutine.  

## Type of change

## Testing

## Documentation

## Final Checklist:

### DIFF
--- a/internal/controllers/app_controller.go
+++ b/internal/controllers/app_controller.go
@@ -443,12 +443,14 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App, logger 
 				}
 			}
 
-			err = r.watchDeployEvents(ctx, app, &wc, wl, &process, r.Recorder)
+			watcher, err := initWatcher(ctx, app, &wc, wl, &process, r.Recorder)
 			if err != nil {
 				return appReconcileResult{
 					err: fmt.Errorf("failed to get deploy events: %w", err),
 				}
 			}
+			go r.watchDeployEvents(ctx, app, &wc, wl, &process, r.Recorder, watcher)
+
 		}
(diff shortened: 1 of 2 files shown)


internal/controllers/app_controller.go
```diff
@@ -443,12 +443,14 @@ func (r *AppReconciler) reconcile(ctx context.Context, app *ketchv1.App, logger
...
 			}
 
-			err = r.watchDeployEvents(ctx, app, &wc, wl, &process, r.Recorder)
+			watcher, err := initWatcher(ctx, app, &wc, wl, &process, r.Recorder)
 			if err != nil {
 				return appReconcileResult{
 					err: fmt.Errorf("failed to get deploy events: %w", err),
 				}
 			}
+			go r.watchDeployEvents(ctx, app, &wc, wl, &process, r.Recorder, watcher)
```
aleksej-paschenko, line 452: we probably want to handle the returned error

stinkyfingers: We create Events for errors inside `watchDeployEvents`. Did you mean something else? If we check the error, it will block the goroutine.

PR #260, opened 2022-06-09
Ketch controller hang

aleksej-paschenko:


aleksej-paschenko (commented):
this approach works for me, shipa-controller doesn't hang and I can redeploy an application

stinkyfingers (approved):

